PR #34, opened 2026-05-08
feat: add CI pipeline with 7 automated checks

### LINKED ISSUE #30: [INFRA] Set up GitHub Actions CI pipeline
## Objective
Add automated CI checks that run on every PR to enforce code 
quality and catch issues before review.

## Tasks
- [ ] Python syntax check on all rule files
- [ ] Rule structure validation (RULE_ID, SEVERITY, FRAMEWORKS etc)
- [ ] Hardcoded credential scan
- [ ] Playbook existence check for every rule
- [ ] Compliance JSON validation
- [ ] API syntax check

## Files to create
- .github/workflows/ci.yml

## Acceptance Criteria
- [ ] CI runs automatically on every PR to dev and main
- [ ] Fails correctly when a rule is missing required fields
- [ ] Passes on the current codebase

## Branch
feat/ci-pipeline

ritiksah141:
## What does this PR do?

Adds a GitHub Actions CI workflow that runs seven automated checks on every pull request to `dev` and `main`. All checks were run locally against the current codebase before pushing — no failures.

## Type of change
- [ ] New scan rule
- [ ] Remediation playbook
- [ ] Bug fix
- [ ] Frontend component
- [ ] API endpoint
- [x] Documentation

Not applicable. This PR adds CI infrastructure, not a scan rule.

## Testing
- [ ] Tested against a real Azure free trial subscription

All seven checks were run locally from the repository root before opening this PR. Full commands are documented in `docs/ci-pipeline.md`.

## What the issue asked for

| Check | Description |
|---|---|
| Python syntax | `py_compile` on all `scanner/rules/az_*.py` files |
| Rule structure | validates `RULE_ID`, `SEVERITY`, `FRAMEWORKS` exist and are correctly typed |
| Hardcoded credentials | grep-based scan for literal secrets, keys, and connection strings |
| Playbook existence | every rule file must have a matching `playbooks/cli/fix_<rule>.sh` |
| Compliance JSON | all three framework JSON files parse correctly and are non-empty |
| API syntax | `py_compile` on all files under `api/` |

## What was added beyond the issue

Four additions were made to improve correctness and reliability:

**RULE_ID uniqueness check (Check 2)**
Two rule files sharing the same `RULE_ID` would silently merge findings in scan reports. The structure check now collects all IDs before reporting and fails if any are duplicated.

**Playbook bash syntax validation (Check 4)**
The original requirement was to check that a playbook file exists. A `.sh` with a bash syntax error crashes when an operator runs it in production. `bash -n` was added to parse each playbook without executing it, catching structural errors such as unclosed `if` blocks and bad redirects.

**SOC 2 framework included in compliance validation (Check 5)**
`soc2.json` was already present in `compliance/frameworks/` but was not being validated. A trailing comma or encoding error in that file would silently break the SOC 2 compliance mapper at runtime. It is now included in the validation pass alongside the three files specified in the issue.

**Compliance vs rule cross-reference (Check 7)**
A new check verifies that every rule ID referenced in the four framework JSONs (`controls` keys) has a matching rule file in `scanner/rules/`. This catches the case where a rule file is deleted without updating the compliance mappings, which would produce a `KeyError` at scan time.

**CI summary panel**
The final step uses `if: always()` and queries each step's outcome via GitHub Actions step IDs. It writes a markdown table to `$GITHUB_STEP_SUMMARY` so reviewers can see which checks passed and which failed directly on the Actions tab, without opening individual step logs.

## Documentation

Full documentation is at `docs/ci-pipeline.md`. It covers:

- How to run all seven checks locally before opening a PR
- Setup instructions including the `pyyaml` dependency added to `requirements.txt`
- Rationale for each testing method chosen (`py_compile` vs linters, `importlib` vs regex, `bash -n` vs existence-only)
- Edge cases handled by each check
- How the CI summary step works
- A lookup table of every failure message with cause and fix

## Acceptance criteria

- [x] CI runs automatically on every PR to `dev` and `main`
- [x] Fails correctly when a rule is missing required fields
- [x] Passes on the current codebase

## Related issue
Closes #30

## Checklist
```

---

## What Could Go Wrong (and How to Fix It)

| Problem | Symptom | Fix |
|---|---|---|
| A rule file doesn't expose `FRAMEWORKS` at module level | Check 2 fails | Add `FRAMEWORKS = {...}` at the top of the rule file |
| `az_stor_003.py` has no matching playbook | Check 4 fails | You created `fix_az_stor_003.sh` in the last issue — verify it's in `playbooks/cli/` |
| Trailing comma in a JSON file | Check 5 fails | Open the JSON and remove the trailing comma before the closing `}` |
| The `api/` directory doesn't exist yet | Check 6 skips gracefully | The workflow handles this — no failure |
| `requirements.txt` doesn't include a needed import | Check 2 or 6 fails at import | Add the missing package to `requirements.txt` |

---

## Understanding the Key Design Decisions

**Why `py_compile` and not `pylint` or `flake8`?**  
`py_compile` only checks syntax — it can't fail on style or linting preferences that differ across contributors. For a hackathon CI, syntax failures are objective; style failures would generate noise and false negatives.

**Why `importlib` for structure validation, not regex?**  
Regex on Python source is fragile — a field could be defined dynamically or in a base class. Actually importing the module and checking `hasattr()` is the only reliable way.

**Why does the credential scan exclude `os.environ` and `os.getenv`?**  
Those patterns are the *correct* way to use environment variables. The scan targets literal assignment (`secret = "abc123"`), not the environment variable lookup pattern.

**Why `set -euo pipefail` is NOT used in the workflow shell steps?**  
GitHub Actions already exits on non-zero by default in `run:` blocks. The explicit `FAIL=0 / exit 1` pattern is used instead for clarity and to collect *all* failures in one pass before exiting.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,370 @@
+name: OpenShield CI
+
+on:
+  pull_request:
+    branches:
+      - dev
+      - main
+
+jobs:
+  ci-checks:
+    name: Run All CI Checks
+    runs-on: ubuntu-latest
+
+    steps:
+      # ── 1. Checkout the code ─────────────────────────────────────────
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      # ── 2. Set up Python ─────────────────────────────────────────────
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      # ── 3. Install dependencies ───────────────────────────────────────
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+
+      # ── CHECK 1: Python syntax on all rule files ───────────────────────
+      - name: Python syntax check (rule files)
+        id: syntax_check
+        run: |
+          echo "=== Checking Python syntax on scanner/rules/ ==="
+          FAIL=0
+          for f in scanner/rules/az_*.py; do
+            if ! python -m py_compile "$f" 2>&1; then
+              echo "SYNTAX ERROR: $f"
+              FAIL=1
+            else
+              echo "OK: $f"
+            fi
+          done
+          if [ "$FAIL" -eq 1 ]; then
+            echo "One or more rule files have syntax errors."
+            exit 1
+          fi
+
+      # ── CHECK 2: Rule structure validation + RULE_ID uniqueness ──────
+      - name: Rule structure validation
+        id: structure_check
+        run: |
+          echo "=== Validating rule file structure ==="
+          python - <<'PYEOF'
+          import os
+          import importlib.util
+          import sys
+          from collections import defaultdict
+
+          rules_dir = "scanner/rules"
+          required_fields = ["RULE_ID", "SEVERITY", "FRAMEWORKS"]
+          valid_severities = {"CRITICAL", "HIGH", "MEDIUM", "LOW", "INFO"}
+          failures = []
+          seen_ids = defaultdict(list)
+
+          for filename in sorted(os.listdir(rules_dir)):
+              if not filename.startswith("az_") or not filename.endswith(".py"):
+                  continue
+
+              filepath = os.path.join(rules_dir, filename)
+              spec = importlib.util.spec_from_file_location("rule", filepath)
+              mod = importlib.util.module_from_spec(spec)
+
+              try:
+                  spec.loader.exec_module(mod)
+              except Exception as e:
+                  failures.append(f"{filename}: import error — {e}")
+                  continue
+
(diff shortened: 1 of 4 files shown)


Vishnu2707:
good work @ritiksah141, nearly there

one thing blocking the merge, Check 5 expects soc2.json to exist but that file hasnt been created yet, every PR will 
fail CI until its there.

either remove soc2.json from the expected_files list in Check 5 for now, or drop an empty stub in this PR, whichever 
is quicker

sort that and ill merge it same day.

Vishnu2707 (approved):
Approved. Issue was fixed in the previous PR, i missed it out, my bad!

PR #117, opened 2026-06-03
Feat/jwt secret prod fail closed

### LINKED ISSUE #116: [SECURITY] Fail closed when JWT_SECRET is missing in production
## Objective

Prevent production deployments from starting with the insecure default JWT secret.

## Why this matters

OpenShield protects sensitive cloud posture data. If JWT_SECRET falls back to a known default, anyone who knows the default can forge valid API tokens for protected /api/* endpoints.

The README warns users to override JWT_SECRET, but the application still starts with change-me-in-production when the environment variable is missing.

## Current behaviour

In api/app.py:

- If JWT_SECRET is not set, the app logs a warning.
- It then uses change-me-in-production.
- This is convenient for local demos but dangerous for deployed environments.

## What to build

Add environment-aware configuration so local development remains easy, but production fails closed.

Recommended behaviour:

- If FLASK_DEBUG=true or OPENSHIELD_ENV=development, allow the default with a clear warning.
- If OPENSHIELD_ENV=production or RENDER=true, require a strong JWT_SECRET and fail startup if missing.
- Add minimum length guidance, for example at least 32 random characters.

## Files likely involved

- api/app.py
- startup.sh
- docs/api-render-deploy.md
- docs/azure-setup.md
- README.md
- tests/

## Acceptance criteria

- [ ] Production startup fails if JWT_SECRET is missing or set to the default
- [ ] Local development can still run with an explicit development mode
- [ ] Tests cover both development and production configuration paths
- [ ] Deployment docs explain how to generate and set a secure JWT secret
- [ ] Smoke tests do not rely on the production-insecure default

ritiksah141:
What does this PR do?
  
 Adds environment-aware JWT secret validation so production deployments refuse to start when JWT_SECRET is missing, set to the known insecure default, or shorter than 32 characters, while keeping local development zero-config.

 Type of change

  - [x] Bug fix
  - [x] API endpoint
  - [x] Documentation

 Rule details (if applicable)

 Not applicable. This is an API security hardening change, not a scan rule.

 Problem

 api/app.py fell back to the hardcoded value change-me-in-production whenever JWT_SECRET was not set in the environment. Anyone who knows this public default can forge valid HS256 tokens and gain full access to all protected /api/* endpoints. The README warned about this but nothing in the application enforced it at startup.

  Solution

  Environment-aware resolution via a new _resolve_jwt_secret() helper in api/app.py.

 Production is detected when OPENSHIELD_ENV=production or RENDER=true (Render sets this automatically on all deployments). In production mode, startup raises a RuntimeError and the process exits if any of the following are true:

  - JWT_SECRET is not set
  - JWT_SECRET equals change-me-in-production
  - JWT_SECRET is shorter than 32 characters

  Development and unspecified environments (no production signal present) allow the default with a loud warning log, keeping local runs and the existing test suite zero-config.

startup.sh now exports OPENSHIELD_ENV=production by default so every gunicorn-based deployment is covered without needing to configure an extra variable. The Render deployment is also covered independently by the automatic RENDER=true signal.

  Files changed

  - api/app.py - added _resolve_jwt_secret(), _is_production(), _is_development(), and _DEFAULT_JWT_SECRET / _MIN_JWT_SECRET_LENGTH constants; replaced the inline
  warn-and-continue block with a single call to _resolve_jwt_secret()
  - startup.sh - exports OPENSHIELD_ENV=production by default
  - tests/smoke_test.py - removed the change-me-in-production fallback; exits with a clear error if JWT_SECRET is not set explicitly
  - tests/test_jwt_config.py - new file; 14 tests covering development permissive paths, unspecified environment regression guard, and all production fail-closed
  conditions via both OPENSHIELD_ENV and RENDER signals
  - .github/workflows/deploy.yml - removed || 'change-me-in-production' fallback from the smoke test step so CI never uses the insecure default
  - README.md - updated security note to describe fail-closed behaviour and minimum length
  - docs/api-render-deploy.md - updated section 2.2 to reflect that the smoke test no longer defaults to the insecure value
  - docs/azure-setup.md - added OPENSHIELD_ENV=production and the generate command to the .env example block

  Generate a strong secret

  python -c "import secrets; print(secrets.token_urlsafe(32))"

  Testing

  - [x] 14 new unit tests in tests/test_jwt_config.py cover both the development permissive path and all production fail-closed conditions; all pass
  - [x] Full test suite (65 tests) passes with no regressions
  - [x] Production fail-closed verified manually: OPENSHIELD_ENV=production with no JWT_SECRET raises RuntimeError and aborts import
  - [x] Local development verified: running without any production signal starts the app with a warning
  - [x] Smoke test (21/21) run against the live Render deployment at https://openshield-api.onrender.com confirms the API is fully operational with the new enforcement
  in place

  Related issue

  Closes #116 

  Checklist

  - [x] My branch name follows the convention: feat/jwt-secret-prod-fail-closed


Vishnu2707 (approved):
Approved.

PR #143, opened 2026-06-14
Fix: Make Flask app test-safe and run full pytest suite in CI

emon22-ts:
## What does this PR do?
Makes the Flask app safe to import without a live DATABASE_URL, then adds the full Python test suite to CI.

## Type of change
- [ ] New scan rule
- [ ] Remediation playbook
- [x] Bug fix
- [ ] API endpoint
- [ ] Documentation
- [ ] Compliance mapping

## Problem
Importing api.app triggered create_app() which immediately called DatabaseManager() and db.run_migrations(). DatabaseManager.__init__ reads os.environ["DATABASE_URL"] and raises KeyError when it is not set. This caused the full test suite to fail at collection time with no DATABASE_URL available in CI.

### api/app.py
Wrapped the migration block in a DATABASE_URL check:
- If DATABASE_URL is set → run migrations as before (production and staging unchanged)
- If DATABASE_URL is not set → log a clear info message and skip migrations (unit tests and local dev without a database)

### .github/workflows/ci.yml
Added CHECK 8 — runs the full pytest suite excluding smoke_test.py with OPENSHIELD_ENV=development so JWT secret validation stays permissive during tests.

## Acceptance criteria
- [x] python3 -m pytest -q runs locally without requiring a production database
- [x] Importing api.app does not trigger migrations when DATABASE_URL is unset
- [x] CI runs all test files under tests/ excluding smoke tests
- [x] Deployment startup still runs migrations when DATABASE_URL is set
- [x] All existing CI checks still pass

## Testing
- [x] Verified locally — app creates successfully without DATABASE_URL
- [x] All CI checks passing locally

## Related issue
Closes #[ISSUE NUMBER]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -337,6 +337,15 @@ jobs:
           echo "=== Running rule regression tests ==="
           pytest tests/test_rules_*.py -v --tb=short
 
+      # ── CHECK 8: Full Python test suite ──────────────────────────────
+      - name: Run Python test suite
+        id: pytest_check
+        run: |
+          echo "=== Running full Python test suite ==="
+          python -m pytest tests/             --ignore=tests/smoke_test.py             -v --tb=short -q
+        env:
+          OPENSHIELD_ENV: "development"
+
       # ── Final summary — always runs, shows per-check pass/fail ────────
       - name: CI Summary
         if: always()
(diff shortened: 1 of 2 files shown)


TFT444 (approved):
Hey @emon22-ts ,

I reviewed this. The api/app.py fix is the right approach, wrapping migrations in the DATABASE_URL check makes the app itself safe to import without a database which is the correct root cause fix. CI is green and the full pytest step in CI is a solid addition on top of the existing rule regression tests.

 Two small things to note: the pytest command has both -v and -q flags which are contradictory, pick one or You can leave as it is.

Also the PR description still says Closes #[ISSUE NUMBER], looks like the actual issue number was not filled in  just put the pr number at the end, please.
 Approving this, good work.

ritiksah141 (approved):

PR #144, opened 2026-06-15
feat(auth): require JWT for GET /api/* endpoints with optional public demo mode

TFT444:
## What does this PR do?

Fixes a security gap where all GET requests to /api/* were publicly accessible without a JWT. Real Azure scan data including findings, compliance status, score, and resources should not be readable without authentication in production.

## Type of change

- [x] Bug fix
- [x] API endpoint
- [x] Documentation

## Changes

**api/app.py** — removed `_is_public_get()` which was making all `/api/*` GET routes public. Replaced with `_ALWAYS_PUBLIC` set containing only `/` and `/health`. Added `OPENSHIELD_PUBLIC_DEMO=true` env var support so deployments running a public demo dashboard can still allow unauthenticated GET requests explicitly. Added `DATABASE_URL` guard so the app can start without a database during tests and local dev.

**tests/test_auth.py** — new test file covering default mode (GET `/api/*` returns 401 without JWT) and demo mode (GET `/api/*` allowed without JWT, POST always requires JWT, `/health` always public).

**.github/workflows/ci.yml** — added `tests/test_auth.py` to the CI pytest command so auth coverage is enforced on every PR.

**docs/api-reference.md** — updated auth section to reflect the new default behavior and added a reference table explaining demo mode vs production mode.

## Acceptance criteria

- [x] `/health` remains public in all environments
- [x] Production `/api/*` GET routes return 401 without a valid JWT by default
- [x] Public demo mode can explicitly allow unauthenticated GET routes via `OPENSHIELD_PUBLIC_DEMO=true`
- [x] Docs explain when JWT is required and when demo mode applies
- [x] Tests cover default production behavior and explicit public demo behavior

## Testing

- [x] Verified locally — app starts without DATABASE_URL
- [x] No hardcoded credentials or secrets
- [x] All CI checks passing

## Checklist

- [x] My code follows the patterns in CONTRIBUTING.md
- [x] I have not committed any real Azure credentials or secrets
- [x] My branch name follows the convention: feat/description

## Related issue

Closes #141

PR #161, opened 2026-07-05
Security hardening: body size limit, Gemini key header, AI rate limiting, playbook path traversal

### LINKED ISSUE #149: [FIX 1] Security hardening: JWT enforcement, API key exposure, request limits, path traversal
## Objective
Fix the five security findings from the codebase audit.

## Findings addressed
- SEC-001 HIGH: All GET /api/* endpoints bypass JWT, add authentication enforcement on GET routes
- SEC-002 HIGH: No request body size limit on AI endpoints, add a maximum payload size
- SEC-003 HIGH: Gemini API key sent as URL query param and appears in logs, move it to a header or request body
- SEC-004 MEDIUM: No rate limiting on AI endpoints, add basic rate limiting
- SEC-005 LOW: Path traversal in playbook file path, sanitise and validate the path

## Acceptance criteria
- [ ] GET /api/* endpoints require valid JWT
- [ ] AI endpoints reject oversized bodies
- [ ] Gemini key no longer appears in any URL or log
- [ ] Rate limiting active on AI endpoints
- [ ] Playbook path traversal blocked
- [ ] CI passes, base branch dev

## Category
Security

TFT444:
## Summary
Closes #149 (FIX 1: Security hardening).

- **SEC-001** (GET /api/* bypasses JWT): already fixed on `dev` via #144; added a regression test to lock it in.
- **SEC-002** (no request body size limit): added `MAX_CONTENT_LENGTH` (2MB) with a JSON 413 error handler.
- **SEC-003** (Gemini key leaks via URL/logs): moved the API key from the `?key=` query param to the `x-goog-api-key` header.
- **SEC-004** (no rate limiting on AI endpoints): added a lightweight in-memory sliding-window limiter (20 req/min/IP), applied to all 5 AI routes; auto-disabled in test mode.
- **SEC-005** (playbook path traversal): `rule_id` is now validated against a strict pattern and the resolved script path is confirmed to stay inside `playbooks/cli/` before being read.

## Test plan
- [x] `tests/test_app_security.py` — JWT enforcement on GET routes, 413 on oversized body
- [x] `tests/test_ai_provider.py` — Gemini key sent via header, absent from URL
- [x] `tests/test_rate_limit.py` — limiter enforces per-route/per-IP limits, bypassed in test mode
- [x] `tests/test_findings_playbook.py` — traversal/absolute-path/slash rule_id rejected, valid rule_id still works
- [x] Full suite: 144 passed, 2 skipped, 2 pre-existing failures unrelated to this change (confirmed identical on a clean `dev` checkout)


m-khan-97 (changes requested):
Reviewed the full diff. The Gemini header fix and the playbook path-traversal fix (regex allowlist + `resolve()` + `is_relative_to()` double-guard) are both correct and well covered by tests — good work on SEC-003 and SEC-005.

Two things on the rate limiter (SEC-004) I want to raise before we call this closed:

1. **State is per-process, but `startup.sh` runs `gunicorn --workers 2`.** `_hits` is a plain module-level dict, so each worker process keeps its own count. That means "20 requests/min per IP" is actually enforced as up to ~40/min with our current startup script — not a hypothetical multi-instance scenario, it's true today.
2. **The limiter keys on `request.remote_addr` with no `ProxyFix`/`X-Forwarded-For` handling**, and we sit behind Render's proxy. If Render doesn't forward the real client IP straight through to the container, every external client collapses into a single `remote_addr`, which turns "20/min per client" into "20/min for the whole API" — a much harsher and almost certainly unintended limit. Can we confirm what Render actually hands us here before relying on this in prod?

Minor/non-blocking: `_hits` entries for a given IP+path are never evicted once that key goes quiet, so there's slow unbounded memory growth over a long-running process. Worth a TODO, not urgent.

None of this is a regression risk and everything else here is solid, but I'd like the rate-limiting behavior confirmed against our actual Render setup (or moved to something process-shared, e.g. Redis-backed) before we treat SEC-004 as fully closed. Requesting changes on that basis — happy to re-review quickly once addressed.

Vishnu2707:
@TFT444 , fixes requested.

ritiksah141 (changes requested):
I reviewed this against #149. The security fixes are moving in the right direction, and the focused tests pass locally, but I am requesting changes before merge.

Findings:

1. The branch fails the current dev lint and format gate. Current dev now enforces Ruff checks, and the touched files do not pass that gate:
   - `api/app.py:263` has `E402 Module level import not at top of file`
   - `api/routes/findings.py:100` has `E741 Ambiguous variable name: l`
   - `api/routes/findings.py:109` has `F541 f-string without placeholders`
   - `ruff format --check` would reformat `api/app.py`, `api/rate_limit.py`, `api/routes/ai.py`, `api/routes/findings.py`, and `api/services/ai_provider.py`

   Please rebase or merge latest `dev`, then run the current lint and format checks before this merges.

2. `api/rate_limit.py:30` keys the AI rate limit on `request.remote_addr`. On Render or any reverse-proxy deployment, that value can be the proxy or load balancer address instead of the real client. That means unrelated users can share one bucket, and one noisy client can rate-limit everyone behind the same proxy. Since SEC-004 is meant to protect AI endpoints per caller, this should use a trusted proxy-aware client identity helper, for example via configured proxy handling or a controlled `access_route`/forwarded-header strategy.

3. This branch is stale against current `dev` and needs revalidation with the newer app changes. The branch predates the current observability and CI updates, and the merge base shows overlap in `api/app.py` and `api/routes/findings.py`. Please update the branch and make sure the current `/ready`, `/metrics`, request-id/error handling, async scan persistence, and new CI checks still pass after the security changes are applied.

Validation I ran:

- Focused security tests: `17 passed`
- Full pytest: `125 passed, 2 skipped, 1 failed`
- The full pytest failure is the known local ONNX Runtime temp-directory error in `tests/test_ai_hallucination_guard.py::TestHallucinationGuard::test_vector_store_purity`, not related to this PR
- `git diff --check dev...HEAD` passed
- PR checks currently shown on GitHub are the older `Run All CI Checks` and CodeQL checks, not the current dev CI pipeline

TFT444:
Gonna be fix soon. 
I was busy with Something 

thanks for the review

m-khan-97 (changes requested):
Re-reviewed after the latest commits. The two things I flagged are genuinely fixed:

- The rate limiter is now Postgres-backed (`pg_advisory_xact_lock` + a `rate_limit_hits` table), so it's correctly shared across all `gunicorn --workers 2` processes instead of counting separately per worker.
- `ProxyFix(x_for=1, x_proto=1)` is added and there's an actual end-to-end test through the real WSGI stack proving `request.remote_addr` resolves from `X-Forwarded-For` instead of Render's proxy address. Good attention to testing this through `app.test_client()` rather than `test_request_context`, which would've missed the bug entirely.

But there's a new problem introduced by the fix itself: **this PR is now in a real merge conflict with `dev`**, and it's not just a git mechanics issue.

`api/models/finding.py` in this PR adds `CREATE TABLE IF NOT EXISTS rate_limit_hits` inside `DatabaseManager.create_tables()`. But #164 (Alembic migrations), merged into `dev` today, deleted `create_tables()` and `run_migrations()` entirely — schema is Alembic-only now, and `init_db()` is a deprecated no-op. GitHub confirms this PR is `CONFLICTING` / `DIRTY` against current `dev` for exactly this reason.

Even setting the conflict aside, resolving it by keeping this hunk wouldn't work: `create_tables()` has no caller left in the app (nothing invokes it at boot anymore), so `rate_limit_hits` would never actually get created in any environment that follows the new Alembic-only deploy process. Every rate-limit check would then throw `relation "rate_limit_hits" does not exist` — which, per this same PR's own fail-open design, gets caught and silently allows the request through. Net effect: SEC-004 would quietly no-op in dev/staging/prod, and nothing in the test suite would catch it, since the rate-limit tests all mock `_check_and_record`/`_get_db` rather than hitting a real schema.

Please rebase onto current `dev` and add a proper Alembic migration for `rate_limit_hits` (`alembic revision -m "add rate_limit_hits table"`) instead of the `create_tables()` hunk, per the process #164 documented in `docs/database-migrations.md`.

Everything else here — the Gemini header fix, the playbook path-traversal guard, the test coverage — still looks correct.

TFT444:
Hey everyone, just wanted to give an update on this PR.

All the feedback has been addressed. The Alembic migration for the rate_limit_hits table is now added properly (revision c7a2e9f1b3d4) so it follows the new migration process from #164. The branch is rebased onto current dev, no merge conflict, lint is clean, and the rate limiter is Postgres-backed with an advisory lock so it works correctly across all gunicorn workers. ProxyFix is also in place so remote_addr resolves the real client IP through Render's proxy.

CI is mostly green 16 out of 17 checks pass. The only failing check is Container Scan (Trivy), but that's a pre-existing vulnerability issue on the dev base image, not something introduced by this PR. This PR doesn't touch the Dockerfile, requirements.txt, or any dependencies, so Trivy would fail the same way on a clean dev checkout. It's being tracked separately.

@Vishnu2707  @m-khan-97 , @ritiksah141 could you take another look when you get a chance? Happy to answer any questions.

Vishnu2707:
@m-khan-97 @ritiksah141 - Requested changes are addressed, kindly approve the PR.

m-khan-97 (approved):

ritiksah141 (approved):

PR #162, opened 2026-07-05
Reliability fixes: score 500, DB pooling, async CVE enrichment, thread-safe rate limit

TFT444:
## Summary
Closes #150 (FIX 2: Reliability).

- **REL-001** (`/api/score` always 500): removed a duplicate `GROUP BY severity` in `get_score()` that caused a SQL syntax error on every call.
- **REL-002** (fragile interval query): replaced the string-substituted `INTERVAL '%s minutes'` with `make_interval(mins => %s)` in `recover_stale_scans()`.
- **REL-003** (new connection per request): `DatabaseManager` now borrows connections from a shared `psycopg2.pool.ThreadedConnectionPool` per DSN (size configurable via `DB_POOL_MAX_CONN`, default 10) instead of opening a new PostgreSQL connection on every request.
- **REL-004** (CVE enrichment times out on large scans): `POST /api/scans/<id>/enrich` now runs enrichment in a background thread and returns 202 immediately; clients poll `GET /api/scans/<id>` (`cve_enrichment_status`) for completion, instead of the request blocking on NVD's ~7s-per-lookup rate limit until it times out.
- **REL-005** (NVD rate-limit guard not thread-safe): `_wait_for_rate_limit()` is now lock-guarded so concurrent callers (e.g. background enrichment threads) can't both read a stale timestamp and fire requests too close together.

### Bug found and fixed during self-review
A failed enrichment write (e.g. in `update_cve_fields`) could leave the DB connection in an aborted-transaction state. The follow-up call to mark the scan `FAILED` reused that same connection *before* returning it to the pool, so it silently raised `InFailedSqlTransaction` and got swallowed — leaving the scan stuck showing `ENRICHING` forever with no way to tell it had actually failed. Fixed by rolling back before that follow-up write, with regression tests (`test_background_enrichment_rolls_back_before_marking_failed`).

### Known trade-offs (not fixed, out of scope for this issue)
- If the whole process crashes mid-enrichment, there's no `recover_stale_scans`-style recovery job for a scan stuck at `ENRICHING` (only scan execution has that today). Worth a follow-up issue if needed.
- No cap on concurrent enrichment threads, but they're naturally serialized by the NVD rate-limit lock anyway.

## Test plan
- [x] `tests/test_database_manager_reliability.py` — single `GROUP BY`, correct score deduction, `make_interval` SQL, pool acquire/return/reuse-per-DSN, broken-connection discard
- [x] `tests/test_scans_enrich.py` — 202 + background thread scheduling, already-completed/in-progress/missing-scan/no-findings branches, success/failure background outcomes, rollback-before-FAILED regression
- [x] `tests/test_nvd_client.py` — new `TestRateLimitThreadSafety` proving concurrent callers are serialized, not interleaved
- [x] `tests/smoke_test.py` — TC-33/34 updated for the async contract, plus a new polling assertion (TC-34b) that verifies enrichment actually reaches `COMPLETED`, not just that it was accepted
- [x] Full suite: 144 passed, 2 skipped, 2 pre-existing failures unrelated to this branch (confirmed identical on a clean `dev` checkout)


Vishnu2707:
@TFT444 , kindlly fix the conflicts please before merging.

Vishnu2707:
@TFT444 , CI is failing, touchbase with this PR pls.

TFT444:
@Vishnu2707 
Only one check is actually failing  Container Scan (Trivy). That's a pre-existing issue on the dev base image and is not related to any changes in this PR. We're not touching the Dockerfile or any dependencies so it would fail identically on a clean dev checkout.

The other two that show as failing Backend Tests and CI Summary both actually passed individually. They appear red on the PR page because when any job in a workflow run fails (Trivy in this case), GitHub marks the entire workflow run as failed and surfaces all jobs under it with a red indicator, even the ones that passed.

So the code is clean, all tests pass, lint is green. The only thing left is for reviewers to re-review and update their review status.

Vishnu2707 (approved):
Approved. All the changes are addressed.

PR #166, opened 2026-07-06
fix(compliance): correct duplicate CIS control mappings, TLS version compare, Graph pagination

TFT444:
## Summary
- Fix AZ-CMP-001 sharing CIS control 7.2 with unrelated AZ-CMP-002 (VM public IP/NSG check vs. disk encryption check)
- De-duplicate 13 rules that shared CIS control_id values across `cis_azure_benchmark.json`, their rule files, `docs/rules-reference.md`, and `website/content.js` — CIS is an enumerated checklist so each control should map to exactly one rule
- Replace lexicographic TLS version comparison in `az_pqc_001` with a numeric parser, fixing the case where `"1.10"` was misread as below `"1.3"`
- Follow `@odata.nextLink` pagination in `get_conditional_access_policies` instead of dropping all results past the first page

## Known limitation
The newly assigned CIS numbers for the 13 de-duplicated rules are a best-effort internal re-mapping based on each rule's own description text; they have not been verified against the official CIS Azure Foundations Benchmark document and should get a compliance SME spot-check.

## Test plan
- [x] New tests: `tests/test_cis_benchmark_mapping.py`, `tests/test_az_pqc_001.py`, `tests/test_graph_pagination.py`
- [x] Merged latest `dev` (which landed a ruff lint/format CI gate) and resolved 3 formatting conflicts, then re-ran `ruff format --check` / `ruff check` — all pass
- [x] Full suite run: 146 passed, 2 skipped, no regressions (2 pre-existing unrelated failures: `test_ai_hallucination_guard.py::test_vector_store_purity`, `test_az_net_008.py::test_non_compliant_no_backend_pool`)
- [x] Self-reviewed via `/code-review --effort high`


Vishnu2707 (approved):
TLS version bug and pagination fix were needed. CIS remapping looks reasonable but needs a proper check against the actual benchmark doc before we go live. approving!

PR #174, opened 2026-07-10
Test: Add missing coverage for DatabaseManager, API routes, Sentinel ingest and RAG pipeline

emon22-ts:
## What does this PR do?
Closes the four test coverage gaps identified in the audit by adding unit tests for DatabaseManager, five API route files, Sentinel HMAC signing and field mappings, and the RAG pipeline.

## Type of change
- [ ] New scan rule
- [ ] Remediation playbook
- [ ] Bug fix
- [ ] API endpoint
- [ ] Documentation
- [ ] Compliance mapping
- [x] CI and Testing

## Coverage gaps addressed
1. CI-001 — DatabaseManager: tests for init, DSN handling, Finding dataclass, SEVERITY_WEIGHTS and get_score() SQL aggregation logic
2. CI-002 — API routes: tests for score, compliance, drift, resources and prioritization endpoints covering 200, 400 and 500 responses, including non-empty success paths for drift (ADDED/REMOVED classification) and resources (risk mapping and by_risk_level counts)
3. CI-003 — Sentinel ingest: tests for HMAC-SHA256 signature correctness, all severity field mappings, missing field defaults, send() retry count (asserts call_count == 3) and early exit on success
4. CI-004 — RAG pipeline: tests for loader.py document loading, chunker.py splitting with exact overlap verification, retriever.py error handling, and build_vectorstore() pipeline (collection creation, add, rename and return count)

## Files added
1.tests/test_database_manager.py - 20 tests
2.tests/test_api_routes.py - 24 tests
3.tests/test_sentinel_ingest.py - 14 tests
4.tests/test_rag_pipeline.py - 24 tests

## Testing
- [x] 86 tests passing locally
- [x] All tests run without a live database or Azure credentials

## Related issue
 Refs #153

### DIFF
--- a/tests/test_database_manager.py
+++ b/tests/test_database_manager.py
@@ -1,0 +1,212 @@
+"""Unit tests for DatabaseManager — connection, CRUD, and scoring logic."""
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from api.models.finding import DatabaseManager, Finding, SEVERITY_WEIGHTS
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _make_db(dsn="postgresql://test:test@localhost/test"):
+    """Return a DatabaseManager with a mocked psycopg2 connection."""
+    with patch("api.models.finding.psycopg2.connect") as mock_connect:
+        db = DatabaseManager(dsn=dsn)
+        mock_conn = MagicMock()
+        mock_cursor = MagicMock()
+        mock_conn.cursor.return_value.__enter__ = MagicMock(return_value=mock_cursor)
+        mock_conn.cursor.return_value.__exit__ = MagicMock(return_value=False)
+        mock_connect.return_value = mock_conn
+        db.conn = mock_conn
+    return db, mock_conn, mock_cursor
+
+
+def _make_finding(**kwargs):
+    defaults = {
+        "rule_id": "AZ-NET-001",
+        "rule_name": "NSG allows broad inbound",
+        "severity": "HIGH",
+        "category": "Network",
+        "resource_id": "/subscriptions/sub-1/rg/rg-1/nsg-1",
+        "resource_name": "nsg-1",
+        "resource_type": "Microsoft.Network/networkSecurityGroups",
+        "description": "NSG allows broad inbound access.",
+        "remediation": "Restrict inbound rules.",
+        "frameworks": {"CIS": "6.1"},
+        "detected_at": "2024-01-01T00:00:00Z",
+    }
+    defaults.update(kwargs)
+    return Finding(**defaults)
+
+
+# ---------------------------------------------------------------------------
+# Initialisation tests
+# ---------------------------------------------------------------------------
+
+
+class TestDatabaseManagerInit:
+    """Tests for DatabaseManager initialisation."""
+
+    def test_init_with_explicit_dsn(self):
+        """DatabaseManager accepts an explicit DSN without reading environ."""
+        db = DatabaseManager(dsn="postgresql://user:pass@localhost/db")
+        assert db.dsn == "postgresql://user:pass@localhost/db"
+
+    def test_init_reads_database_url_from_env(self, monkeypatch):
+        """DatabaseManager reads DATABASE_URL when no DSN is provided."""
+        monkeypatch.setenv("DATABASE_URL", "postgresql://env:env@localhost/envdb")
+        db = DatabaseManager()
+        assert db.dsn == "postgresql://env:env@localhost/envdb"
+
+    def test_init_raises_without_dsn_or_env(self, monkeypatch):
+        """DatabaseManager raises KeyError when DATABASE_URL is unset."""
+        monkeypatch.delenv("DATABASE_URL", raising=False)
+        with pytest.raises(KeyError):
+            DatabaseManager()
+
+    def test_conn_is_none_before_connect(self):
+        """Connection is None before connect() is called."""
+        db = DatabaseManager(dsn="postgresql://test:test@localhost/test")
+        assert db.conn is None
+
+
+# ---------------------------------------------------------------------------
+# SEVERITY_WEIGHTS tests
+# ---------------------------------------------------------------------------
+
+
+class TestSeverityWeights:
+    """Tests for SEVERITY_WEIGHTS constant."""
+
+    def test_high_outweighs_medium(self):
+        assert SEVERITY_WEIGHTS["HIGH"] > SEVERITY_WEIGHTS["MEDIUM"]
+
+    def test_medium_outweighs_low(self):
+        assert SEVERITY_WEIGHTS["MEDIUM"] > SEVERITY_WEIGHTS["LOW"]
+
+    def test_info_has_zero_weight(self):
+        assert SEVERITY_WEIGHTS.get("INFO", 0) == 0
+
+    def test_all_weights_non_negative(self):
+        for key, value in SEVERITY_WEIGHTS.items():
+            assert value >= 0, f"{key} has negative weight"
+
+
+# ---------------------------------------------------------------------------
+# Finding dataclass tests
+# ---------------------------------------------------------------------------
+
+
+class TestFindingDataclass:
+    """Tests for the Finding dataclass."""
+
+    def test_finding_creation_with_required_fields(self):
+        """Finding can be created with required fields."""
+        finding = _make_finding()
+        assert finding.rule_id == "AZ-NET-001"
+        assert finding.severity == "HIGH"
+
+    def test_finding_to_dict_returns_dict(self):
+        """to_dict() returns a dictionary."""
+        finding = _make_finding()
+        result = finding.to_dict()
+        assert isinstance(result, dict)
+
+    def test_finding_to_dict_contains_rule_id(self):
+        """to_dict() includes rule_id."""
+        finding = _make_finding()
+        result = finding.to_dict()
+        assert result["rule_id"] == "AZ-NET-001"
+
+    def test_finding_optional_fields_default_to_none(self):
+        """Optional fields default to None or empty."""
+        finding = _make_finding()
+        assert finding.scan_id is None
+        assert finding.playbook is None
+        assert finding.id is None
+
+    def test_finding_metadata_defaults_to_empty_dict(self):
+        """metadata defaults to empty dict."""
+        finding = _make_finding()
+        assert finding.metadata == {}
+
+    def test_finding_cve_references_defaults_to_empty_list(self):
+        """cve_references defaults to empty list."""
+        finding = _make_finding()
+        assert finding.cve_references == []
+
+    def test_finding_exploit_available_defaults_to_false(self):
+        """exploit_available defaults to False."""
+        finding = _make_finding()
+        assert finding.exploit_available is False
+
+
+# ---------------------------------------------------------------------------
+# Score calculation tests
+# ---------------------------------------------------------------------------
+
+
+# ---------------------------------------------------------------------------
+# Score calculation tests — call real get_score() via mock cursor
+# ---------------------------------------------------------------------------
+
+
+class TestScoreCalculation:
+    """Tests for DatabaseManager.get_score() SQL aggregation and floor logic."""
+
+    def _make_db_with_score(self, fetchone_value):
+        db = DatabaseManager(dsn="postgresql://test:test@localhost/test")
+        mock_conn = MagicMock()
+        mock_cursor = MagicMock()
+        mock_cursor.__enter__ = MagicMock(return_value=mock_cursor)
+        mock_cursor.__exit__ = MagicMock(return_value=False)
+        mock_cursor.fetchall.return_value = fetchone_value
+        mock_conn.cursor.return_value = mock_cursor
+        mock_conn.closed = False
+        db.conn = mock_conn
+        db._get_conn = MagicMock(return_value=mock_conn)
+        return db
+
+    def test_get_score_returns_100_with_no_findings(self):
+        """get_score() returns 100 when there are no findings."""
+        db = self._make_db_with_score([])
+        assert db.get_score() == 100
+
+    def test_get_score_deducts_high_severity(self):
+        """get_score() deducts 10 per HIGH finding via SQL aggregation."""
+        db = self._make_db_with_score([("HIGH", 2)])
+        assert db.get_score() == 80
+
+    def test_get_score_deducts_multiple_severities(self):
+        """get_score() deducts correct amounts for mixed severities."""
+        db = self._make_db_with_score([("HIGH", 1), ("MEDIUM", 2)])
+        assert db.get_score() == 80
+
+    def test_get_score_floors_at_zero(self):
+        """get_score() floors at 0 — never returns negative."""
+        db = self._make_db_with_score([("HIGH", 20)])
+        assert db.get_score() == 0
+
+    def test_get_score_returns_integer(self):
+        """get_score() returns an integer."""
+        db = self._make_db_with_score([("LOW", 1)])
+        assert isinstance(db.get_score(), int)
+
+    def test_get_score_calls_execute(self):
+        """get_score() calls cursor.execute — SQL actually runs."""
+        db = self._make_db_with_score([])
+        db.get_score()
+        db.conn.cursor().__enter__().execute.assert_called_once()
+
+    def test_severity_weights_high_outweighs_medium(self):
+        assert SEVERITY_WEIGHTS["HIGH"] > SEVERITY_WEIGHTS["MEDIUM"]
+
+    def test_severity_weights_medium_outweighs_low(self):
+        assert SEVERITY_WEIGHTS["MEDIUM"] > SEVERITY_WEIGHTS["LOW"]
+
+    def test_severity_weights_info_is_zero(self):
+        assert SEVERITY_WEIGHTS.get("INFO", 0) == 0
--- a/tests/test_database_manager.py
+++ b/tests/test_database_manager.py
@@ -1,6 +1,6 @@
 """Unit tests for DatabaseManager — connection, CRUD, and scoring logic."""
 
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock
 
 import pytest
 
@@ -10,19 +10,6 @@ from api.models.finding import DatabaseManager, Finding, SEVERITY_WEIGHTS
 # ---------------------------------------------------------------------------
 # Helpers
 # ---------------------------------------------------------------------------
-
-
-def _make_db(dsn="postgresql://test:test@localhost/test"):
-    """Return a DatabaseManager with a mocked psycopg2 connection."""
-    with patch("api.models.finding.psycopg2.connect") as mock_connect:
-        db = DatabaseManager(dsn=dsn)
-        mock_conn = MagicMock()
-        mock_cursor = MagicMock()
-        mock_conn.cursor.return_value.__enter__ = MagicMock(return_value=mock_cursor)
-        mock_conn.cursor.return_value.__exit__ = MagicMock(return_value=False)
-        mock_connect.return_value = mock_conn
-        db.conn = mock_conn
-    return db, mock_conn, mock_cursor
 
 
 def _make_finding(**kwargs):
--- a/tests/test_database_manager.py
+++ b/tests/test_database_manager.py
@@ -4,7 +4,8 @@ from unittest.mock import MagicMock
 
 import pytest
 
-from api.models.finding import DatabaseManager, Finding, SEVERITY_WEIGHTS
+from api.models.finding import DatabaseManager, Finding
+from openshield.severity import SEVERITY_WEIGHTS
 
 
 # ---------------------------------------------------------------------------
--- a/tests/test_database_manager.py
+++ b/tests/test_database_manager.py
@@ -188,12 +188,3 @@ class TestScoreCalculation:
         db = self._make_db_with_score([])
         db.get_score()
         db.conn.cursor().__enter__().execute.assert_called_once()
-
-    def test_severity_weights_high_outweighs_medium(self):
-        assert SEVERITY_WEIGHTS["HIGH"] > SEVERITY_WEIGHTS["MEDIUM"]
-
-    def test_severity_weights_medium_outweighs_low(self):
-        assert SEVERITY_WEIGHTS["MEDIUM"] > SEVERITY_WEIGHTS["LOW"]
-
-    def test_severity_weights_info_is_zero(self):
-        assert SEVERITY_WEIGHTS.get("INFO", 0) == 0
--- a/tests/test_rag_pipeline.py
+++ b/tests/test_rag_pipeline.py
@@ -1,0 +1,313 @@
+"""Unit tests for the RAG pipeline — ai/loader.py, ai/chunker.py, ai/retriever.py."""
+
+from pathlib import Path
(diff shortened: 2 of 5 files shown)


Vishnu2707:
@parthrohit22 , @ritiksah141  - assigning the review to u both, seems like you are the code owners and do know the codebase well and what value add this brings in, please do touch base with this.

Vishnu2707:
Container scan - Broken CI, the push for the fix has been done to dev which triggers in the next commit, hence go ahead without minding that CI response.

tests/test_database_manager.py
```diff
@@ -0,0 +1,187 @@
...
+
+# ---------------------------------------------------------------------------
+# Score calculation tests
+# ---------------------------------------------------------------------------
+
+
+class TestScoreCalculation:
+    """Tests for score calculation logic."""
+
+    def test_score_starts_at_100(self):
```
ritiksah141: These tests recreate the scoring arithmetic inside the test instead of calling DatabaseManager.get_score(). A regression in the query aggregation or the production floor logic would still pass. Please mock the connection rows and assert results returned by db.get_score(). Also add coverage for at least one persistence method or narrow the stated CRUD scope, since _make_db is currently unused.

tests/test_sentinel_ingest.py
```diff
@@ -0,0 +1,213 @@
...
+def test_send_returns_false_after_retries():
+    """send() returns False after 3 failed attempts."""
+    ingest = _make_ingest()
+    mock_response = MagicMock()
+    mock_response.status_code = 500
+    mock_response.text = "Internal Server Error"
+    with patch("requests.post", return_value=mock_response):
+        with patch("time.sleep"):
+            result = ingest.send([ingest.normalise(RAW_FINDING, "scan-001")])
+    assert result is False
```
ritiksah141, line 204: This only verifies the final False result. It would still pass if send() stopped retrying after one request. Keep the requests.post mock and assert call_count == 3. The exception path should make the same assertion so the claimed retry behavior is protected.

tests/test_rag_pipeline.py
```diff
@@ -0,0 +1,210 @@
...
+        assert len(chunks) == 1
+
+    def test_long_document_produces_multiple_chunks(self):
+        """A document longer than chunk_size produces multiple chunks."""
+        from ai.chunker import chunk_documents
+
+        long_content = "word " * 300
+        docs = [self._make_doc(long_content)]
+        chunks = chunk_documents(docs, chunk_size=100, chunk_overlap=10)
+        assert len(chunks) > 1
```
ritiksah141, line 129: The PR says chunk overlap is covered, but this assertion only proves that a long document creates multiple chunks. Please assert that adjacent chunk contents contain the requested overlap, or adjust the stated coverage scope.

ritiksah141 (changes requested):
I pulled the branch and reviewed it against the current architecture and latest dev. All 72 added tests pass, ruff passes, and the branch has no merge conflicts. The earlier Trivy failure is not part of this review, as requested in the PR discussion.

I found three coverage gaps that should be corrected before approval:

1. The score tests reproduce arithmetic locally instead of executing DatabaseManager.get_score(), so regressions in production SQL aggregation and floor handling would not be detected.
2. The Sentinel retry tests only assert the final False result and do not prove that three attempts occurred.
3. The RAG chunking tests pass a chunk_overlap value but never verify overlap behavior.

Please also update the PR description because Closes #174 refers to this pull request itself rather than a related issue.

m-khan-97 (changes requested):
Went through this against the actual source for the three areas Ritik flagged, since I wanted to confirm they're real before piling on rather than just trusting the summary. All three check out:

1. **Score tests never call the real code path.** `TestScoreCalculation` in `test_database_manager.py` reproduces the arithmetic inline (`score -= SEVERITY_WEIGHTS["HIGH"]`) instead of calling `DatabaseManager.get_score()`. A bug in the actual SQL aggregation/floor logic would sail through untouched.
2. **Sentinel retry tests don't verify retry count.** I checked `sentinel/ingest.py:64` — `send()` genuinely does `for attempt in range(1, 4)` (3 attempts). But `test_send_returns_false_after_retries` and `test_send_retries_on_exception` only assert the final `False`, never `mock_post.call_count == 3`. Change that loop to `range(1, 2)` and both tests would still pass.
3. **Chunk overlap is asserted by count, not content.** Confirmed in `ai/chunker.py:51` — `chunk_overlap` directly controls `start = split_pos - chunk_overlap` for the next chunk. None of the chunker tests check that consecutive chunks actually share the expected overlapping text; they only check chunk count and `chunk_index` presence.

Agreeing with Ritik's request for changes on this basis — the coverage numbers are real (72 tests, all passing), but for the specific claims this PR makes (CI-001 through CI-004 "closed"), these three gaps mean the corresponding regressions could land undetected. Worth fixing before merge rather than as a fast-follow, since the whole point of this PR is closing coverage gaps precisely so regressions get caught.

One separate housekeeping item, also from Ritik's review and still true: the PR description says "Closes #174" but #174 is this PR's own tracking issue for the coverage gaps — please point it at the actual issue number if there is one, or drop the auto-close keyword if #174 is just this PR.

Nice test for the HMAC signature itself, by the way — `test_build_signature_uses_hmac_sha256` independently recomputes the expected signature and compares against the real function's output rather than mocking it away. That's the right pattern; would be good to see the retry-count and chunk-overlap tests follow the same approach.

Vishnu2707:
@emon22-ts , there are quite some changes to be done, please do touch base with this PR.

parthrohit22 (changes requested):
I reviewed PR #174 against its four changed test files, the existing review threads, and the underlying testing requirements in issue #153.

I agree with @ritiksah141’s three unresolved findings, which @m-khan-97 also independently confirmed:

1. The score tests reproduce the arithmetic instead of calling `DatabaseManager.get_score()`.
2. The Sentinel tests assert only the final `False` result and do not prove that all three retry attempts occur.
3. The chunker tests verify the number of chunks but not the configured overlap between adjacent chunks.

I will not repeat those inline comments. I found two additional coverage gaps:

- Issue #153 explicitly identifies `ai/embed.py` as part of the untested RAG pipeline, but this PR does not test `build_vectorstore()` or the vector-store creation path.
- The drift, resources, and prioritization success tests use empty database results and assert only `200`/JSON. Their main comparison, aggregation, ranking, and response-mapping behavior is therefore not exercised.

The DatabaseManager concern is particularly important: this PR’s current head contains a duplicated `GROUP BY severity` inside `get_score()`, while all the new tests still pass because none execute the real method. This defect is inherited from the older base rather than introduced by this PR, and updating the branch from current `dev` should incorporate its correction. A real `get_score()` test should nevertheless be added to prevent recurrence.

The HMAC-SHA256 test is well designed and independently verifies the production function. Most CI checks are green, and I am not attributing the acknowledged Trivy failure to this PR.

Before approval, please:

- Address the three existing unresolved review threads.
- Add non-empty success-path tests for drift, resources, and prioritization.
- Add mocked coverage for `ai/embed.py`.
- Update the branch from `dev` and rerun CI.
- Correct the related issue reference: `Closes #174` currently refers to this pull request itself; the matching OpenShield testing issue is #153.
- Correct the per-file test counts to 20 DatabaseManager, 20 API route, 13 Sentinel, and 19 RAG tests, which total the stated 72.

Requesting changes until the claimed coverage gaps and the underlying acceptance criteria are fully addressed.

TFT444 (commented):
@emon22-ts fix the following changes. they asking for

Vishnu2707 (commented):
This PR is in an idle state, the contributor hasn't responded yet.

parthrohit22:
> This PR is in an idle state, the contributor hasn't responded yet.

I'll takeover from the same branch remotely , expect updated pr within 12 hours, Thank you.

emon22-ts:
Hi i will do the changes that are required and really sorry i was busy with my personal stuff

emon22-ts:
Hi, all review feedback addressed:

1. Score tests now call real DatabaseManager.get_score() via mocked cursor returning severity/count rows — SQL aggregation and floor logic are properly exercised
2. Sentinel retry tests now assert mock_post.call_count == 3 proving all 3 attempts occur, plus added test_send_stops_after_first_success verifying early exit on 200
3. Chunker overlap tests now verify actual shared content between adjacent chunks, plus added test_no_overlap_when_zero
4.Added TestEmbedPipeline covering build_vectorstore() — raises without chromadb, raises with no documents, and calls load_all_documents
5. PR description updated — Closes #153 (the actual issue), not #174

80 tests passing, ruff check and ruff format both clean.

Ready for re-review. Thank you! @Vishnu2707 @parthrohit22 @ritiksah141 @TFT444 @m-khan-97

parthrohit22 (changes requested):
Thanks for addressing the earlier feedback. I re-reviewed commit d1a2c3d against the current test files and production implementations.

The DatabaseManager.get_score() tests now execute the real method with mocked query rows, and the Sentinel retry tests correctly verify three attempts and early exit after success. Those points are resolved.

A few blockers remain before approval:

1. Update the branch from dev. The PR is currently 26 commits behind, and its inherited get_score() SQL still contains a duplicated GROUP BY severity. Current dev already contains the correction. Please rebase or merge dev and rerun CI.
2. Add non-empty success-path coverage for drift, resources and prioritization. These tests remain unchanged: they use empty query results and assert only 200/JSON. They do not exercise drift comparison, resource aggregation/risk mapping, prioritization scoring, sorting or response construction.
3. Make the overlap assertion deterministic. The new overlap test uses repetitive content and accepts any matching suffix, so it can pass even with chunk_overlap=0. Use unique input and assert the exact configured overlap between adjacent chunks.
4. Strengthen the build_vectorstore() success test. It currently catches and ignores all exceptions and only asserts load_all_documents() was called. Remove the catch-all and verify collection creation, add() calls, final rename through modify(), and the returned chunk count.
5. Update the PR description. It still references Closes #174 and retains the old test counts. The related testing issue is #153; since it is already closed, Refs #153 would be clearer.

Requesting changes until these items are addressed. The current CI run is green, but it must be rerun after synchronising the branch with dev.

emon22-ts:
hey @parthrohit22 it will be done soon in next 12 hours .

emon22-ts:
Hi @parthrohit22 @ritiksah141 , all review feedback addressed:

1.Fixed prioritization test mocks — _make_prioritization_db now uses a single cursor with fetchone.side_effect for the two sequential fetchone() calls (scan_id, then total count), matching the route's actual single-cursor query pattern
 
2.Added missing resource_name field to high_rule/low_rule fixtures in test_prioritization_high_severity_ranked_first, which the route reads unconditionally

3.Fixed Python 3.9 compatibility crash in cbom.py — added from __future__ import annotations to support the Dict[str, Any] | None union syntax (PEP 604), which was breaking create_app() for the entire test suite

4.Fixed lint failures — added trailing newline to test_api_routes.py (W292) and ran ruff format on cbom.py and test_api_routes.py

24 tests passing, ruff check and ruff format both clean, all CI jobs green.
Ready for re-review. Thank you!

parthrohit22 (dismissed):
Reviewed the PR end-to-end — commits, changed files, and mapped everything back to issue [#153](https://github.com/openshield-org/openshield/issues/153). All four coverage gaps (CI-001 → CI-004) are addressed, the tests were verified against the actual source (get_score() weights/floor, build_signature() HMAC-SHA256, normalise() severity map + defaults, send() 3-attempt retry), the cbom.py from __future__ import annotations is a legitimate and necessary Python 3.9 compat fix, and CI is fully green (19/19).

Thanks a lot @emon22-ts  for turning this around and addressing every piece of review feedback — the score tests now exercise the real DatabaseManager.get_score() via a mocked cursor instead of re-implementing the arithmetic, the sentinel retry tests assert call_count == 3 (plus the early-exit-on-success case), and the chunker tests now verify actual shared overlap content between adjacent chunks. All three of the earlier review threads are genuinely resolved in code. 🙌

A couple of non-blocking nits I'll defer to @Vishnu2707  as maintainer, since the merge decision is yours:

_make_db() in test_database_manager.py is still unused — either wire it into a persistence-method test or drop it.
The test_database_manager.py docstring mentions "CRUD" but there's no actual CRUD/persistence-method test — worth trimming the docstring or adding one small insert/save test to match scope.
Neither of these blocks the merge in my view — happy to see them handled in a follow-up if you'd prefer to keep this one moving. @Vishnu2707  over to you on whether to merge as-is or fold the nits in first.

parthrohit22 (changes requested):
Thanks for sticking with this @emon22-ts — I know it's been open a while. Went through
the current head properly and the three original threads are genuinely closed, not just
papered over. Score tests hit the real `get_score()`, retry tests assert `call_count == 3`
with the early-exit case, and the overlap test checks actual shared content. The
`build_vectorstore` rewrite is much better too — real assertions instead of the catch-all.

The thing still blocking is what @Vishnu2707 raised back on the 25th: drift and resources
never got their non-empty tests. Prioritization did, those two didn't. Every drift test
runs with zero or one scan and every resources test runs with no scan, so the comparison
and aggregation logic never actually executes. I swapped `added_keys` and `removed_keys`
in the drift route and scrambled the `rank_to_risk` map in resources — both times the full
suite still passed, 24/24. So those routes could return inverted output and nothing here
would notice.

Should be a small fix: one drift test with two scans that differ so ADDED/REMOVED gets
classified, and one resources test with mixed-severity rows so the risk mapping and
`by_risk_level` counts get checked.

A few smaller things while you're in there:

- `cbom.py` is the only non-test file in the PR, and the Python 3.9 reasoning doesn't hold
  up — CI and the Dockerfile are both on 3.11, and that union syntax works natively from
  3.10. It's harmless but it'll conflict on rebase since `dev` doesn't have it. Either drop
  it or split it out.
- Branch is 3 commits behind `dev` — worth syncing and letting CI rerun.
- PR description still says `Closes #174`, which points at this PR itself. Should be
  `Refs #153`. Counts are stale too — it's 86 tests now, not 72 (24/24/14/24).
- Still open from earlier: `_make_db()` at the top of `test_database_manager.py` is unused,
  and the docstring mentions CRUD but there's no persistence test.

Good work overall, the HMAC test especially — recomputing the signature independently
instead of mocking it is exactly right. Get drift and resources covered and I think this
is done. Merge call is @Vishnu2707's.


For reference, here's what I checked: pulled the current head and ran the four files —
84 passed, 2 skipped (the skips are chromadb missing locally and correctly guarded).
`ruff check` and `ruff format --check` both clean across 160 files. Rather than just
confirming the tests pass, I mutated the production code to see whether they'd actually
catch a regression: removed the floor in `get_score()`, dropped the Sentinel retry loop
from 3 attempts to 1, and made the chunker ignore `chunk_overlap` — each one failed the
right test, so those fixes are real. Ran the same exercise on drift and resources, and
neither mutation was caught, which is what confirmed the gap above. Also checked the
cbom change against the repo's actual Python target and verified `dev`'s version imports
fine without it.

Vishnu2707:
@emon22-ts , please address the drift test mentioned in the previous conversations, rest LGTM.

emon22-ts:
Ok sir , I am on it

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

emon22-ts:
@Vishnu2707 @parthrohit22 Hi, all remaining feedback addressed:

1. Drift - added test_drift_with_two_scans_classifies_added_removed: two scans with differing findings so ADDED/REMOVED classification actually executes. Mutating event_type or swapping added/removed keys will now fail this test.

2. Resources - added test_resources_with_mixed_severity_rows: rows with risk_rank 3/2/1 so rank_to_risk mapping and by_risk_level counts are verified. Scrambling rank_to_risk will now fail this test.

3. Removed unused _make_db helper from test_database_manager.py

4. cbom.py dropped from this PR (no longer in diff)

5. Rebased from latest dev

6. PR description updated  - Refs #153, 86 tests (24 sentinel + 24 RAG + 14 DB + 24 routes)

Ready for final review. Thank you for sticking with this!

TFT444 (changes requested):
Thanks for the Aug 18 push @emon22-ts, the resources test now properly pins the risk mapping and `by_risk_level` counts. Almost there, but a few things still need attention before merge:

1. **`test_api_routes.py` (drift):** `test_drift_with_two_scans_classifies_added_removed` only asserts `summary["total"] > 0`. Swapping `added_keys`/`removed_keys` in the route (the mutation from the last review) still passes. Please assert `summary["added"] == 1`, `summary["removed"] == 1`, and that the AZ-NET-001 event is `type == "ADDED"`.

2. **`test_sentinel_ingest.py`:** `test_normalise_unknown_severity_defaults_to_zero` claims unknown severities fall back to MEDIUM, but it just passes `"MEDIUM"` and asserts 2. The real behavior (`ingest.py:76`) is that unknown severity raises `ValidationError`. Test that, or drop it.

3. **`cbom.py`:** the `from __future__ import annotations` change was already asked to be dropped or split out; it's still here.

Nits, same push: the `if len(rankings) >= 2:` guard in the prioritization test makes it silently pass, and the 3 `SEVERITY_WEIGHTS` tests at the bottom of `TestScoreCalculation` duplicate `TestSeverityWeights`.

Fix 1-3 and this is good to go from my side.

emon22-ts:
Hi @Vishnu2707 @parthrohit22 @m-khan-97 @TFT444 , all items addressed plus fixes for the OWASP repo changes:

1. Drift test — now asserts summary["added"]==1, summary["removed"]==1, event type=="ADDED"/"REMOVED" and rule_violated=="AZ-NET-001" for the ADDED event
2. Sentinel unknown severity — renamed to test_normalise_unknown_severity_raises_validation_error, now passes "UNKNOWN" and verifies ValidationError is raised
3.  cbom.py — removed from diff, restored to upstream
4. Prioritization silent guard — replaced with hard assertion (len(rankings) >= 2)
5. Duplicate SEVERITY_WEIGHTS tests removed from TestScoreCalculation

Also fixed for OWASP repo changes:
- SEVERITY_WEIGHTS moved to openshield.severity — import updated
- ai/retriever.py rewritten to BM25 — retriever tests updated accordingly
- ai/embed.py rewritten to BM25 — embed tests updated (no chromadb dependency)
- Prioritization mock fixed for new fetchall side_effect pattern for rules + severity counts
- Sentinel tests updated for mandatory severity validation in OWASP ingest.py — test_normalise_missing_fields_use_defaults and test_normalise_generates_timestamp_when_missing now pass a valid severity field

71 tests passing, ruff clean. Ready for final review. Thank you.

parthrohit22 (changes requested):
Thank you @emon22-ts for the updates and for addressing the earlier coverage feedback.

Problem with current approach

One issue remains: `test_normalise_unknown_severity_raises_validation_error` passes whether `normalise()` raises or not. The broad exception handler and `or raised` condition mean it does not protect the invalid-severity validation path.

Correct approach

Use `pytest.raises(ValidationError, match="severity")` around the `normalise()` call and remove the broad exception handling.

Implementation steps

Please make this assertion explicit and rerun the backend test suite.

Risks / edge cases

Without this, a regression that accepts or silently remaps an unknown severity can pass CI and affect Sentinel finding risk accuracy.

Verified: the current GitHub checks are green, including backend tests, CodeQL, dependency review, and security scans.

emon22-ts:
Hi @Vishnu2707 @parthrohit22 @m-khan-97 @TFT444 
 Fixed - test_normalise_unknown_severity_raises_validation_error now uses pytest.raises(ValidationError, match="severity") directly with no broad exception handling. A regression that silently accepts or remaps an unknown severity will now fail CI.

Ready for final review. Thank you.

TFT444 (approved):
Re-reviewed. Both fixes are in: test_normalise_unknown_severity_raises_validation_error now uses pytest.raises(ValidationError, match='severity') correctly, and the drift test asserts exact added == 1 / removed == 1 values with specific rule IDs. Approved.

parthrohit22 (changes requested):
Thanks for the thorough work on expanding coverage, and for addressing the earlier review feedback. The Sentinel retry and validation assertions, along with the drift/resource scenarios, are much stronger now.

I found a few remaining items that need to be addressed before I can approve:

1. In `test_rag_pipeline.py`, `test_build_vectorstore_calls_load_all_documents` catches every exception and only patches `VECTORSTORE_DIR`. The implementation writes through `INDEX_PATH` using `Path.write_text()`, so the current mocks do not isolate the write. Please patch both paths to a temporary directory and remove the broad exception handler so an index-write failure cannot pass silently.

2. The compliance route tests currently only assert a `200` response, and the shared mock returns a CIS payload for every framework. Please assert that `get_compliance_score()` receives the requested framework and validate the corresponding response payload for every supported framework.

3. `api/routes/cbom.py` is still changed, although this PR is intended to be test-only and the description states that the CBOM change was removed. Please revert this unrelated import.

Once these are addressed and the backend test suite is rerun, I’m happy to re-review.

emon22-ts:
It will be done by today . Thanks

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

TFT444 (changes requested):
@emon22-ts three blockers from the Sep 18 review remain unaddressed:

1. `test_build_vectorstore_calls_load_all_documents` has a broad catch-all that can hide index-write failures. Patch both `VECTORSTORE_DIR` and `INDEX_PATH` to a temp dir and remove the catch-all.
2. Compliance route tests only assert `200`. Please assert that `get_compliance_score()` receives the correct framework argument and validate the response payload for each supported framework.
3. `api/routes/cbom.py` still contains an unrelated `from __future__ import annotations` change. Please remove it or split it into a separate PR.

m-khan-97 (approved):
I re-reviewed the current head against the three outstanding blockers. They are now resolved: the BM25 build test patches both output paths, performs the real write, and no longer swallows exceptions; all six supported compliance routes assert the exact framework passed to the database layer and validate their payloads; and api/routes/cbom.py is no longer in the PR diff. I ran the four focused test files locally: all 88 tests passed. Approved.

TFT444 (approved):
All blockers addressed in 2c245cc: vectorstore test now patches both VECTORSTORE_DIR and INDEX_PATH, compliance route tests assert per-framework payloads, and the unrelated cbom.py import is removed. Verified by m-khan-97 as well. Approving.

TFT444:
@parthrohit22 
Can you please rereview it quickly? It's all good; I believe I'm just waiting for your request changes to change

parthrohit22 (approved):
Re-reviewed `2c245cc`. All three items from my 18 Sep review are properly addressed — approving.

**1. `test_build_vectorstore_calls_load_all_documents`** — now patches both `VECTORSTORE_DIR` and `INDEX_PATH` to a real `tempfile.TemporaryDirectory`, runs `build_vectorstore()` for real, and asserts `bm25_index.json` actually exists afterwards. The broad `except` is gone, so an index-write failure now fails the test instead of passing silently. `test_build_vectorstore_returns_chunk_count` got the same treatment, which I didn't ask for but is the right call for consistency.

**2. Compliance route tests** — `_check_framework()` asserts `get_compliance_score.assert_called_once_with(framework)`, so a route that ignored the path parameter and always queried CIS would now fail. It also echoes the framework back from the payload and checks `passed + failed == total_controls`, across all six supported frameworks. The added `error` → 500 case alongside the existing 400/500 paths rounds out the contract.

**3. `api/routes/cbom.py`** — `git diff upstream/dev <head> -- api/routes/cbom.py` is empty. The unrelated change is gone and the PR is test-only again.

CI is green across all 21 checks. Thanks for staying with this through a long review cycle — the coverage here is real coverage, not line-count coverage, and the mutation-resistance in the drift and resource tests is what makes it worth merging.

ritiksah141 (approved):

PR #176, opened 2026-07-12
infra 7 : Terraform for Render/Vercel + GitHub OIDC for Azure credentials

TFT444:
## Summary

Closes #160.

- `infra/terraform/`: Terraform for the *current* live topology — one Render web service (`openshield-api`), one Render Postgres (`openshield-db`), a shared secrets env group, and two Vercel projects (`frontend` dashboard, `website` docs site). Every credential-shaped value is a `sensitive = true` Terraform variable with no default — nothing is applied, no live credentials exist in this repo. See `infra/terraform/README.md` for the one-time setup and plan/apply flow.
- `.github/workflows/terraform-plan.yml`: runs `terraform fmt`/`validate` on every PR touching `infra/terraform/**`; also runs `terraform plan` once `TF_API_TOKEN` is configured (gracefully skipped until then, workflow still passes).
- `deploy.yml`: Azure smoke-test authentication switched from a stored `AZURE_CLIENT_SECRET` to GitHub OIDC federation via `azure/login@v3`. No application code changes needed — `DefaultAzureCredential()` already picks up workload-identity federation automatically.
- `docs/ci-oidc-setup.md`: the one-time `az ad app federated-credential create` commands needed on the existing `openshield-scanner` service principal (I have no Azure access to run these myself).
- `docs/secrets-inventory.md`: every credential-shaped value in the project and where it lives (GitHub secret vs Render env var vs Terraform variable).

## Known scope limits (called out explicitly, not silently swept under)

- **#160 depends on #157** (service topology), which isn't merged yet — PR #172 is in progress and adds its own `render.yaml` blueprint. This PR models today's live topology only; reconciling Terraform vs the `render.yaml` blueprint is a follow-up once #172 lands (documented in `infra/terraform/README.md`).
- **No live infra access.** I have no Render/Vercel/Azure/Terraform Cloud credentials, so `terraform plan` has never run against real infrastructure. Verified instead via `terraform fmt -check` and `terraform validate` (both pass, provider schemas confirmed against each provider's own docs). The plan tier variables (`render_postgres_plan`, `render_web_service_plan`) are documented best-guesses from `docs/api-render-deploy.md` — flagged inline as needing confirmation against the real dashboard before the first apply.
- **`AZURE_CLIENT_SECRET` is not deleted** — that's a manual, deliberate step for whoever holds repo admin access, done only after confirming OIDC actually works in a real `deploy.yml` run (see `docs/ci-oidc-setup.md`).
- **OIDC scope**: applies only to `deploy.yml`'s maintainer-owned smoke-test Azure identity, not to how the product itself authenticates to end users' own Azure subscriptions (those are user-supplied credentials; OIDC federation is inherent to GitHub Actions and can't apply there).

## Security/quality pass

- `gitleaks` scan of every new commit: 0 leaks.
- `trivy config` scan of `infra/terraform/`: 0 misconfigurations.
- All 13 credential-shaped Terraform variables are `sensitive = true` with no defaults; confirmed `AZURE_CLIENT_SECRET` appears nowhere under `infra/`.
- New/modified workflow permissions are minimal (`terraform-plan.yml`: `contents: read, pull-requests: write`; `deploy.yml`: added only `id-token: write, contents: read`).
- `outputs.tf` exposes only non-sensitive identifiers/URLs, never `connection_info` or env var values.

## Test plan

- [x] `terraform fmt -check -recursive` — clean
- [x] `terraform validate` (local backend override, no live credentials) — `Success! The configuration is valid.`
- [x] Both new/modified workflow YAML files parse correctly
- [x] `ruff check` / `ruff format --check` — clean (this PR touches no `.py` files)
- [x] Full pytest suite: 260 passed, same 2 pre-existing unrelated local failures, zero regressions
- [x] `gitleaks` + `trivy config` — clean

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,6 +3,10 @@ name: Deploy API to Render
 on:
   workflow_dispatch:   # allows manual trigger from GitHub UI
 
+permissions:
+  id-token: write
+  contents: read
+
 jobs:
   deploy:
     name: Deploy to Render
@@ -85,6 +89,19 @@ jobs:
             exit 0
           fi
 
+      # ── Azure OIDC login ──────────────────────────────────────────────
+      # Exchanges this workflow run's GitHub OIDC token for a short-lived Azure
+      # token via the openshield-scanner service principal's federated
+      # credential — no AZURE_CLIENT_SECRET needed. See docs/ci-oidc-setup.md
+      # for the one-time Azure AD setup this depends on.
+      - name: Azure login (OIDC)
+        if: steps.check_config.outputs.is_configured == 'true' || github.event_name == 'workflow_dispatch'
+        uses: azure/login@v3
+        with:
+          client-id: ${{ secrets.AZURE_CLIENT_ID }}
+          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+
       # ── Smoke tests ────────────────────────────────────────────────────
       - name: Run smoke tests against live deployment
         if: steps.check_config.outputs.is_configured == 'true' || github.event_name == 'workflow_dispatch'
@@ -93,7 +110,6 @@ jobs:
           JWT_SECRET: ${{ secrets.JWT_SECRET }}
           AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
           AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
-          AZURE_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
           AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
           RUN_REAL_SCAN: "true"
         run: |
--- a/.github/workflows/terraform-plan.yml
+++ b/.github/workflows/terraform-plan.yml
@@ -1,0 +1,62 @@
+name: Terraform Plan
+
+on:
+  pull_request:
+    paths:
+      - "infra/terraform/**"
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  plan:
+    name: Terraform fmt / validate / plan
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: infra/terraform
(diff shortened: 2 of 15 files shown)


m-khan-97 (approved):
Reviewed the full diff — this is exceptionally well-scoped for a "we have zero live infra access" constraint. A few things stood out as particularly well-reasoned:

- **Correctly limits OIDC to CI, not runtime.** `AZURE_CLIENT_SECRET` is deliberately kept out of the Terraform-managed env group and left as a manual Render dashboard entry, not deleted. That's the right call — Render's compute has no Azure workload-identity federation, so the long-running API/worker service still needs a standing credential; only the ephemeral GitHub Actions runner in `deploy.yml` can use OIDC. Good to see that distinction actually reflected in the code, not just asserted in prose.
- **Honest about the danger zone.** The README is explicit that the Render web service, Postgres instance, and both Vercel projects already exist and must be `terraform import`-ed before any real `terraform apply`, or it'll attempt to create duplicates. That's the single most important thing whoever does the first real apply needs to not skip — worth restating out loud here since it's easy to gloss over in a large README.
- Full CI is green, including `terraform fmt`/`validate` actually running (not just claimed), and the known overlap with #172's `render.yaml` blueprint is flagged as an explicit unresolved follow-up rather than silently left for someone to discover later.

Two minor, non-blocking notes:

1. `docs/ci-oidc-setup.md` and `docs/secrets-inventory.md` describe the Azure identity on the Render web service as distinct from "end-user scan credentials" / "the product feature itself," implying a BYOC flow. As far as I can tell from the current codebase, there's only one server-side Azure service principal in play (`AZURE_SUBSCRIPTION_ID`/`CLIENT_ID`/`TENANT_ID` env vars used by both the API's own scans and the smoke test) — there's no per-user credential flow today. Worth softening that language slightly so it doesn't read as describing a feature that doesn't exist yet.
2. The planning doc under `docs/superpowers/specs/2026-07-12-terraform-oidc-design.md` references `azure/login@v2` in a couple of spots while the shipped workflow and the paired plan doc both correctly use `v3`. Harmless since it's scratch/planning material, not user-facing docs, but a quick pass would keep it internally consistent.

Neither blocks this. Approving — nice work handling a "no credentials, be honest about it" constraint without pretending the risk isn't there.

Vishnu2707 (approved):
Approved, as this addresses the clear scope of the work.

PR #183, opened 2026-07-13
fix(security): resolve all high/critical CodeQL findings (#177, #179, #180)

TFT444:
## Summary

Fixes #177, #179, #180 — all high and critical CodeQL findings (34 total across the 4 issues filed from the code-scanning review; this PR covers the 3 grouped as high/critical, per team decision to split high+critical into one PR and medium into a separate one).

- **#179 (critical — stored XSS)**: `toEmbedUrl()` in `website/script.js` validated embed URLs with a substring check (`raw.includes('youtube.com/embed')`) and returned the **unvalidated raw input** when it matched — bypassable with a crafted string, breaking out of the iframe's `src="..."` attribute. Rewritten to validate via `new URL(raw).hostname` against an exact allowlist. Added `website/test_toEmbedUrl.mjs`, a dependency-free test that loads the real production file (no duplicated logic) and verifies the exploit string is rejected while legitimate YouTube/Vimeo URLs still work.
- **#180 (high — clear-text storage, also a live bug)**: `frontend/src/utils/aiApi.js`'s AI provider API key storage. A prior attempt to fix this same CodeQL alert had removed the `localStorage.setItem` call but left the getter/`isConfigured()` reading from the now-permanently-empty key — **silently breaking all four AI features** regardless of what key a user entered. Moved the key to a module-scoped in-memory variable: satisfies CodeQL (never touches browser storage) and restores the feature. Added `frontend/src/utils/aiApi.test.mjs`.
- **#177 (medium-severity CodeQL, bundled here per the high+critical grouping)**: 27 findings across `api/app.py` and 8 route modules under `api/routes/` where exceptions were reflected back to the client via `str(exc)` or an f-string, including from broad `except Exception` blocks that could surface raw DB driver errors. Replaced with fixed, generic messages; existing/added `logger.error(...)` calls preserve full detail server-side. Extracted a shared `_ai_error_response()` helper in `ai.py` for 4 near-identical handlers. Also fixed an f-string leak in `compliance.py`'s `FileNotFoundError` handler that a plain `"detail"` key search would have missed.

## Self-review

Ran a high-effort, 6-agent parallel review against the full diff. One real issue found in my own work: `escapeHTML()` had been added at 3 interpolation sites in `website/script.js` that turned out to be `.textContent` assignments, not `.innerHTML` — `.textContent` never interprets HTML, so escaping there added no security value while risking visible text corruption (a literal `&amp;` on screen) if the underlying value ever contained a special character. Reverted all 3; `toEmbedUrl()`'s URL-origin validation is the real, load-bearing fix and doesn't depend on them.

Everything else the review surfaced was low-severity/architectural (no shared error-response helper across all 9 API files yet, no repo-wide lint rule preventing a future `str(exc)` regression, two new JS test files each hand-roll similar sandboxing boilerplate since neither `website/` nor `frontend/` has an existing test framework) — noted as follow-up-worthy but not blocking this fix.

## Test plan

- [x] `ruff check` / `ruff format --check` — clean
- [x] Full pytest suite: 348 passed, same 2 pre-existing unrelated local failures, no regressions
- [x] 14 new regression tests (`tests/test_error_exposure.py`, `tests/test_ai_error_exposure.py`) asserting error responses never contain a `"detail"` key or the raw exception text, across every route touched
- [x] `website/test_toEmbedUrl.mjs` (13 cases) and `frontend/src/utils/aiApi.test.mjs` (6 cases) — both load and test the real production source, wired into CI (`ci.yml`: new `website` job, new step in the `frontend` job)
- [x] No merge conflicts against `dev`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -548,6 +548,9 @@ jobs:
       - name: Lint
         run: npm run lint
 
+      - name: Run aiSettings tests
+        run: node src/utils/aiApi.test.mjs
+
       - name: Build
         run: npm run build
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -551,6 +551,21 @@ jobs:
       - name: Build
         run: npm run build
 
+  website:
+    name: Website (script tests)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - name: Set up Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: "22"
+
+      - name: Run website script tests
+        run: node website/test_toEmbedUrl.mjs
+
   # ── Enforce branch flow: main may only receive PRs from dev ───────────────
   # No-op on dev PRs (step skipped -> job succeeds). To allow hotfixes straight
   # to main, widen the condition below to also accept a 'hotfix/*' head branch.
@@ -575,7 +590,7 @@ jobs:
     name: CI Summary
     runs-on: ubuntu-latest
     needs:
-      [lint, rule-validation, secret-scan, sast-bandit, sca-pip-audit, sbom, container-scan, backend-tests, frontend, enforce-source-branch]
+      [lint, rule-validation, secret-scan, sast-bandit, sca-pip-audit, sbom, container-scan, backend-tests, frontend, website, enforce-source-branch]
     if: always()
     steps:
       - name: Build summary
@@ -589,6 +604,7 @@ jobs:
           CONTAINER_SCAN: ${{ needs.container-scan.result }}
           BACKEND_TESTS: ${{ needs.backend-tests.result }}
           FRONTEND: ${{ needs.frontend.result }}
+          WEBSITE: ${{ needs.website.result }}
           ENFORCE_SOURCE: ${{ needs.enforce-source-branch.result }}
         run: |
           python3 - <<'PYEOF'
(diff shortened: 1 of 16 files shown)


m-khan-97 (approved):
Thorough review of all three fixes here, checked against actual source rather than just the description:

**#179 (stored XSS)** — verified the new \`toEmbedUrl()\` logic directly: \`new URL(raw).hostname\` against an exact-match allowlist correctly defeats every bypass class I'd try — substring tricks (\`evil.com/?x=youtube.com/embed\`), subdomain spoofing (\`youtube.com.evil.com\`), and userinfo tricks (WHATWG URL parsing puts anything before \`@\` into userinfo, not hostname, so that vector is closed by construction, not just by the allowlist). The test file backs this with the exact bypass strings you'd expect an attacker to try, including markup injection. Good fix.

**#180 (clear-text storage + the live bug it caused)** — really glad this called out that the *previous* attempted fix had already broken all four AI features by removing the localStorage write but leaving the getter reading from it. Moving to a module-scoped in-memory variable satisfies CodeQL and restores the feature in one change, with tests for both properties (never touches localStorage under any key, and still round-trips correctly in memory). The documented trade-off (key doesn't survive reload) is the right call and clearly explained in the comment.

**#177 (error exposure)** — the sweep across all 8 route files plus \`api/app.py\`'s error handlers is complete and consistent. Good catch on \`compliance.py\`'s f-string leak specifically — that's a real, distinct leak vector from the \`"detail"\` key pattern the rest of the routes had, and a naive grep for \`"detail"\` would've missed it. The 25 new regression tests use a distinctive marker string rather than checking for generic wording, which is the right way to make this kind of test actually catch a regression instead of coincidentally passing.

Also appreciated the self-review note about reverting the \`.textContent\` escaping — recognizing that fix added no security value and could visibly corrupt text is exactly the kind of check that should happen before merge, not after someone notices garbled output in production.

Approving.

ritiksah141 (changes requested):
Two security issues need to be addressed before this PR can be approved.

1. `toEmbedUrl()` validates with `new URL(raw)` but returns the original `raw` string. An allowed hostname can therefore retain attribute-breaking characters. For example, `https://youtube.com/embed/x" onload="alert(1)` passes the protocol and hostname checks and is returned unchanged. Return the canonical `parsed.href` after validating the expected YouTube or Vimeo embed pathname, and add regression cases for quotes or markup characters on an allowed hostname.

2. The in-memory API key change does not remove a legacy `ai_api_key` already stored by earlier releases. Existing users can therefore retain the clear-text secret in localStorage indefinitely unless they manually use `clear()`. Remove the legacy entry during module initialization or a one-time migration, and add a test that pre-populates `ai_api_key` and verifies it is deleted without being read into memory.

The remaining changes and tests look sound. Local verification passed: 329 tests passed, 3 skipped; the 17 targeted error-exposure tests passed; both JavaScript regression suites passed; Ruff and formatting checks passed.

PR #184, opened 2026-07-13
fix(security): drop credential key_id from AZ-IDN-006 debug logging (#178)

TFT444:
## Summary

Fixes #178 — the medium-priority CodeQL batch (2 findings, both the same log statement in `scanner/rules/az_idn_006.py`).

`app_id`/`key_id` are Azure AD identifiers — Microsoft Graph never returns the actual password-credential secret text after creation, so these are GUIDs referencing a credential, not the credential itself. This is very likely CodeQL's naming-heuristic flagging variables near a "credential"/`cred` context rather than a genuine secret leak.

Regardless of true/false-positive status, `key_id` isn't needed to diagnose a malformed `endDateTime` value, so the fix removes it from the log line entirely rather than arguing the classification. `key_id` itself is untouched everywhere else in the file (it's still used correctly in the finding's `resource_id` and `metadata`).

## Test plan

- [x] Added `test_idn_006_malformed_end_date_time_does_not_log_key_id`, which feeds a malformed `endDateTime` and asserts a sentinel `keyId` value never appears in the captured log output
- [x] Verified this test actually catches the bug: temporarily reverted the fix and confirmed the new test fails against the old code, then confirmed it passes against the fix
- [x] `ruff check` / `ruff format --check` — clean
- [x] Full pytest suite: 332 passed, same 2 pre-existing unrelated local failures, no regressions
- [x] No merge conflicts against `dev`

### DIFF
--- a/scanner/rules/az_idn_006.py
+++ b/scanner/rules/az_idn_006.py
@@ -87,9 +87,8 @@ def scan(azure_client: Any, subscription_id: str) -> List[Dict[str, Any]]:
                     already_expired = end_dt < now
                 except ValueError:
                     logger.debug(
-                        "AZ-IDN-006: Invalid endDateTime for app_id=%s key_id=%s: %r",
+                        "AZ-IDN-006: Invalid endDateTime for app_id=%s: %r",
                         app_id,
-                        key_id,
                         end_dt_str,
                     )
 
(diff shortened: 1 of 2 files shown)


m-khan-97 (approved):
Small, clean, well-verified fix. Confirmed the removed \`key_id\` was genuinely unnecessary for diagnosing a malformed \`endDateTime\` — it's still used correctly everywhere else in the file (resource_id, metadata), only pulled from this one debug log line. Like that the PR description doesn't argue the true/false-positive classification and just removes the unneeded field regardless — the right call either way. Test is solid too: it actually reverted the fix locally to confirm the new test fails against the old code before confirming it passes against the fix, which is exactly how you prove a regression test has teeth. Approving.

ritiksah141 (approved):

PR #185, opened 2026-07-13
fix(security): resolve Semgrep SAST findings

TFT444:
## Summary

Resolves the Semgrep SAST findings tracked in #182 across GitHub Actions workflows, backend code, container configuration, and the public website.

## Security remediation

- Prevented workflow command injection by passing dispatch inputs through environment variables before shell use.
- Replaced both `curl | sh` Syft installers with pinned release archives and SHA-256 verification.
- Pinned reported GitHub Actions dependencies to immutable commit SHAs.
- Added Dependabot cooldown periods to reduce exposure to newly published malicious or unstable packages.
- Removed JWT exception details from logs.
- Replaced dynamically assembled SQL with a fixed parameterized query.
- Enforced HTTPS and exact-host validation for NVD and Render API requests; scanner suppressions are limited to these validated calls.
- Hardened Docker Compose services with read-only root filesystems, temporary writable mounts, and `no-new-privileges`.
- Removed JavaScript `eval()` from RSS generation and replaced it with a direct module export.
- Pinned external browser dependencies and added SHA-384 Subresource Integrity checks.

## Acceptance criteria

- [x] All reported HIGH findings resolved.
- [x] All reproduced MEDIUM findings resolved or limited by explicit validation.
- [x] LOW findings reviewed and remediated where reproduced.
- [x] Full local Semgrep scan reports 0 findings.
- [x] Semgrep Cloud scan passes.
- [x] Ruff, Bandit, CodeQL, Trivy, dependency review, secret scanning, backend tests, frontend checks, and CI Summary pass.
- [x] Changes submitted through PR review against `dev`.

## Validation

- Full Semgrep registry scan: **0 findings** across 314 tracked files and 604 rules.
- Semgrep Cloud: **passed**.
- Backend regression tests: **53 passed** locally; full backend CI passed.
- All 11 YAML files parsed successfully.
- Website JavaScript syntax validation passed.

Closes #182.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,16 +5,22 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
     open-pull-requests-limit: 0
 
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
     open-pull-requests-limit: 0
 
   - package-ecosystem: "npm"
     directory: "/frontend"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
     open-pull-requests-limit: 0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -431,7 +431,16 @@ jobs:
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
 
       - name: Install syft
-        run: curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin v1.46.0
+        env:
+          SYFT_VERSION: "1.46.0"
+          SYFT_SHA256: d654f678b709eb53c393d38519d5ed7d2e57205529404018614cfefa0fb2b5ca
+        run: |
+          SYFT_ARCHIVE="syft_${SYFT_VERSION}_linux_amd64.tar.gz"
+          curl --fail --silent --show-error --location \
+            --output "$SYFT_ARCHIVE" \
+            "https://github.com/anchore/syft/releases/download/v${SYFT_VERSION}/${SYFT_ARCHIVE}"
+          echo "${SYFT_SHA256}  ${SYFT_ARCHIVE}" | sha256sum --check --strict
+          sudo tar --extract --gzip --file "$SYFT_ARCHIVE" --directory /usr/local/bin syft
 
       - name: Generate SBOM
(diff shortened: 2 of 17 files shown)


m-khan-97 (approved):
Broad PR, went through each category of fix rather than skimming:

- **SQL fix in \`get_findings()\`** — verified the parameterization is correct: 7 \`%s\` placeholders in the SQL match the 7-element \`params\` tuple positionally, and the \`(%s IS NULL OR col = %s)\` pattern correctly no-ops for unset filters while filtering correctly when set. Checked the one caller that builds a variable filter dict (\`findings.py\`'s \`list_findings\`) — it only ever includes a key when the query param was actually present, so the dict never contains an explicit \`None\`, meaning behavior is unchanged for every real call path. Good fix for what was a real Semgrep-flagged pattern even though the old version wasn't actually exploitable (values were already parameterized, just the clause *structure* was dynamically assembled).
- **SSRF guards on \`nvd_client.py\`/\`render_deploy.py\`** — correct and appropriately narrow (exact scheme/hostname/port match against the one legitimate endpoint each function calls), with a real test (\`test_open_rejects_untrusted_endpoint\`) proving the guard actually fires rather than just existing.
- **Workflow command injection fix, action SHA-pinning, Syft curl\|sh → checksum-verified download, SRI hashes on the website's CDN scripts** — all correct, standard hardening, nothing to flag.
- **\`eval()\` removal in \`generate_rss.js\`** — cleaner besides being safer; \`module.exports\` on \`content.js\` is the right minimal change to make that possible.

One thing worth a direct answer rather than a silent approve: the Docker Compose hardening (\`read_only: true\`, \`no-new-privileges\`, tmpfs mounts) is applied to \`db\` and \`frontend\` but not \`api\` — which is the service actually holding \`JWT_SECRET\`/\`ANTHROPIC_API_KEY\`/\`DATABASE_URL\` and facing untrusted network input, so arguably the one most worth hardening. I checked the Dockerfile and don't see anything that would need root-filesystem writes at runtime (gunicorn, no local file writes), so \`api\` becoming read-only too seems safe — was it left out deliberately, or is this a follow-up? Separately, \`db\`'s data directory is already on the \`postgres_data\` named volume so \`read_only: true\` there should be safe, but I don't see a \`docker compose up\` smoke test anywhere in CI or the PR's validation section confirming the hardened containers actually still start cleanly (particularly Postgres, which is the classic case where read-only-root silently breaks at first checkpoint/WAL write rather than at startup). Worth a quick manual \`docker compose up\` check before merge, or a fast follow-up if that's already been done outside what's captured here.

Neither blocks this — nothing here affects production (Render, not docker-compose, is what's actually deployed) and every check including Semgrep Cloud is green. Approving.

PR #188, opened 2026-07-14
fix(security): complete #183 — XSS output encoding + legacy AI key purge

m-khan-97:
Follow-up to #183, addressing two residual issues @ritiksah141 flagged in review after it merged. Both are real; neither warranted reverting #183, so this patches forward.

## 1. XSS — `toEmbedUrl()` returned the raw input, not the validated URL
`website/script.js`'s embed fallback correctly validated the **host** via `new URL(raw)` against an allowlist, but then returned `raw` — the unmodified input. On an **allowed** host, an attribute-injection payload still passes:

```
https://www.youtube.com/embed/abc" onload="alert(1)
```

`hostname` is `www.youtube.com` (passes the allowlist), but the literal `"` survives and breaks out of the iframe `src="..."` attribute when interpolated. The earlier "rejected" tests only covered *disallowed hosts*, so this class was never exercised.

**Fix:** return `parsed.href` instead of `raw`. The WHATWG URL serializer percent-encodes quotes, spaces, and angle brackets, so the returned value is safe to interpolate. Legitimate embed URLs are unaffected (they contain no characters that get encoded).

## 2. Legacy plaintext key never purged (#180 regression)
The in-memory-key switch removed the `localStorage.setItem('ai_api_key', ...)` write but never removed keys **older builds had already persisted**. Any user who configured a key before that build keeps a plaintext `ai_api_key` in localStorage indefinitely — the exact leak #180 set out to close.

**Fix:** `localStorage.removeItem('ai_api_key')` once on module load (deliberately *not* read into memory) and again in `clear()` as defence-in-depth.

## Verification
- Both regression tests were confirmed to **fail against the current `dev` code** before the fix and **pass after** (not just added green).
- `node website/test_toEmbedUrl.mjs` — passes (2 new attribute-injection-on-allowed-host cases)
- `node frontend/src/utils/aiApi.test.mjs` — passes (3 new legacy-key migration cases)
- `npm run lint` — 0 errors (4 pre-existing warnings in untouched files)

## Note
This is the miss I flagged in the #183 post-merge review: my approval checked the host-validation gate but not the value flowing out of it. These tests close that specific gap. cc @ritiksah141 @Vishnu2707

### DIFF
--- a/frontend/src/utils/aiApi.js
+++ b/frontend/src/utils/aiApi.js
@@ -24,6 +24,13 @@ const TIMEOUT = 30000;
 // persist in localStorage as before.
 let apiKeyInMemory = '';
 
+// One-time migration: builds before the in-memory switch (issue #180) stored
+// the key at localStorage['ai_api_key']. Purge it on load so existing users
+// don't keep a plaintext key sitting in storage indefinitely. Deliberately not
+// read into apiKeyInMemory — the whole point is that the secret no longer lives
+// in a persistent, XSS-readable store.
+localStorage.removeItem('ai_api_key');
+
 export const aiSettings = {
   getProvider: () => localStorage.getItem('ai_provider') || 'anthropic',
   getApiKey:   () => apiKeyInMemory,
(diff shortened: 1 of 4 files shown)


Vishnu2707 (approved):
Thanks for the fixes. LGTM!

PR #197, opened 2026-07-15
feat(scanner): add enterprise App Registration and Managed Identity rules

TFT444:
## What does this PR do?

Adds a cohesive enterprise identity security pack for Microsoft Entra App Registrations and Azure Managed Identities, while consolidating App Registration inventory so all related rules share one cached, paginated Graph request per scan.

## Rules

| Rule | Severity | Detection |
|---|---|---|
| `AZ-IDN-010` | MEDIUM | App Registration has no owner |
| `AZ-IDN-011` | HIGH | Non-loopback HTTP redirect URI |
| `AZ-IDN-012` | MEDIUM | OAuth implicit token issuance enabled |
| `AZ-IDN-013` | MEDIUM | Password credentials present |
| `AZ-IDN-014` | HIGH | Multi-tenant app lacks full application-instance property lock |
| `AZ-IDN-015` | HIGH | Managed identity has Owner or Contributor at subscription scope |

## Existing rule improvement

`AZ-IDN-006` now reuses `AzureClient.get_applications()` instead of performing its own Graph request. The accessor:

- follows `@odata.nextLink`;
- caches successful, empty, and failed inventory states;
- expands only minimal owner IDs;
- requests only security-relevant application properties;
- returns `None` on Graph failure so rules cannot mistake an incomplete scan for compliance.

Managed Identity detection uses Graph's immutable `servicePrincipalType=ManagedIdentity` filter and correlates principal IDs with cached subscription RBAC assignments. Privileged roles are matched by immutable built-in role-definition GUIDs and exact subscription scope.

## Data minimization

Findings do not include:

- tokens or credential values;
- credential key IDs or hints;
- complete redirect URIs;
- owner names, email addresses, or user principal names.

Only counts, boolean states, role names, principal IDs, and required resource identifiers are retained.

## Assignment restriction decision

The proposed `AZ-IDN-016` resource-provider assignment-restriction rule is intentionally deferred. Microsoft documents the new portal feature, but the current public Managed Identity REST response reliably exposes `isolationScope`, not the configured resource-provider restriction collection. This PR does not infer or invent an undocumented field.

## Permissions

Read-only scanning requires:

- Microsoft Graph application permission `Application.Read.All`;
- Azure action `Microsoft.Authorization/roleAssignments/read`.

No Graph write permission is requested by the scanner.

## Compliance

Mappings were added for the repository's NIST CSF 1.1, ISO/IEC 27001:2013, and SOC 2 schemas. The current CIS Azure Foundations 2.0.0 file has no direct controls for these checks, so mappings are explicitly numbered `TBD-IDN-*` placeholders rather than claiming unsupported CIS coverage.

## Testing

- [x] 409 tests passed from a clean checkout
- [x] 3 existing environment-dependent tests skipped
- [x] 58 focused identity tests passed
- [x] Pagination and per-scan caching tested
- [x] Graph and Azure RBAC failures tested as indeterminate states
- [x] Loopback IPv4, IPv6, localhost, HTTPS, and native redirect cases tested
- [x] Exact subscription versus resource-group scope tested
- [x] Ruff lint and format checks passed
- [x] Bash syntax checks passed for all playbooks
- [x] Bandit passed at the CI threshold
- [x] No new dependency introduced
- [ ] Live tenant validation (requires an approved non-production Entra/Azure environment)

## Related issue

Closes #193

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -203,6 +203,9 @@ Use the existing wrapper methods in `scanner/azure_client.py` rather than constr
 | `azure_client.get_sql_server_auditing_policy(resource_group, server_name)` | ServerBlobAuditingPolicy or None |
 | `azure_client.get_key_vaults()` | List of Key Vault objects |
 | `azure_client.get_managed_clusters()` | List of AKS ManagedCluster objects, or `None` on API failure |
+| `azure_client.get_applications()` | Paginated App Registration dictionaries, or `None` on Graph failure |
+| `azure_client.get_managed_identity_service_principals()` | Managed Identity service principals, or `None` on Graph failure |
+| `azure_client.get_subscription_role_assignments()` | Subscription RBAC assignments, or `None` on API failure |
 | `azure_client.get_service_principals()` | List of role assignments for service principals |
 | `azure_client.get_conditional_access_policies()` | List of Conditional Access policy dicts from Microsoft Graph |
 
--- a/compliance/frameworks/cis_azure_benchmark.json
+++ b/compliance/frameworks/cis_azure_benchmark.json
@@ -259,32 +259,32 @@
       "description": "Microsoft recommends a managed node OS upgrade channel for timely security patches. This check has no direct control in the repository's CIS Azure Foundations 2.0.0 benchmark."
     },
     "AZ-IDN-010": {
-      "control_id": "N/A-IDN-010",
+      "control_id": "TBD-IDN-010",
       "control_name": "App Registration ownership (not mapped in CIS Azure Foundations 2.0.0)",
       "description": "Microsoft recommends accountable App Registration ownership. This check has no direct control in the repository's CIS Azure Foundations 2.0.0 benchmark."
     },
     "AZ-IDN-011": {
-      "control_id": "N/A-IDN-011",
+      "control_id": "TBD-IDN-011",
       "control_name": "App Registration redirect URI security (not mapped in CIS Azure Foundations 2.0.0)",
       "description": "Microsoft requires secure redirect URI handling. This check has no direct control in the repository's CIS Azure Foundations 2.0.0 benchmark."
     },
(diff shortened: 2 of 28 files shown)


m-khan-97 (approved):
Reviewed thoroughly — fetched the branch and ran the tests locally (34 passed), plus checked the redirect-URI classifier against edge cases directly.

Strong work:
- **`get_service_principals()` refactor preserves behavior.** It now delegates to the cached `get_subscription_role_assignments()`, and still returns `[]` on failure (when assignments is `None`), so no regression for existing callers. The `_UNSET` sentinel correctly distinguishes "not yet fetched" from "fetched and failed (None)".
- **AZ-IDN-006 correctly reuses the shared `get_applications()`** instead of its own inline Graph fetch, and handles the `None` (inventory-unavailable) case rather than silently treating it as "no findings."
- **Redirect-URI logic (AZ-IDN-011) is correct.** Verified `http://[::1]`, `http://127.0.0.1:8080`, and `http://localhost` are all treated as loopback (not flagged), while `http://evil.com` is flagged and `msalclient://` (non-http scheme) is ignored. `ipaddress.ip_address()` on the bracket-stripped IPv6 host works as intended.
- **Data minimization is enforced and tested** — findings carry counts/booleans only, and the tests explicitly assert the raw URI and a sentinel `keyId` never appear in metadata. Good follow-through on the #178 lesson.
- Every rule has a matching confirmation-gated playbook; the destructive ones (secret deletion, implicit-grant disable) are print-only or `APPLY`-gated.

Minor cosmetic note, non-blocking: the new rules use `TBD-IDN-0xx` as the CIS control_id, while the existing AKS rules used `N/A-AKS-006`. Since the descriptions say these genuinely aren't in CIS Foundations 2.0.0, `N/A-` reads as more accurate than `TBD-` (which implies "coming soon"). Not worth a round-trip.

Approving.

Issue #31, opened 2026-05-06
[DOCS] Update architecture.md to reflect all 20 current rules


Vishnu2707:
## What to build
Update docs/architecture.md to reflect the current state of the 
scanner. The architecture doc currently only lists the original 
10 rules. The project now has 20 rules across 6 categories.

## Changes needed
1. Update the rule modules section to list all current rules
2. Add a rules summary table showing count per category
3. Update the AzureClient methods table in docs/adding-a-rule.md 
   to include any new methods added since the original 10 rules

## Files to update
- docs/architecture.md
- docs/adding-a-rule.md

## How to get started
1. Fork the repo, create branch docs/update-architecture
2. Read docs/architecture.md
3. Compare against scanner/rules/ folder for what is missing
4. Open a PR to dev

## Acceptance Criteria
- [ ] All 20 rules listed
- [ ] Rule count table accurate per category
- [ ] AzureClient methods table up to date

Vishnu2707:
The issue has been already addressed, so closing the issue.

PR #12, opened 2026-05-02
feat: Sentinel integration — ingest.py, 4 KQL rules, setup guide

TFT444:
Closes #4

Wires OpenShield scanner findings into Microsoft Sentinel via Log Analytics and deploys 4 KQL detection rules.

## What was built
- sentinel/ingest.py — pushes findings to Log Analytics via HMAC-SHA256 signed REST API
- sentinel/rules/high_severity_finding.kql — fires on any HIGH/CRITICAL finding
- sentinel/rules/misconfiguration_wave.kql — fires when 5+ HIGHs in single scan (wave score 120 in testing)
- sentinel/rules/persistent_misconfiguration.kql — fires when same resource flagged across 3+ scans
- sentinel/rules/new_resource_type_critical.kql — fires on unknown resource type + CRITICAL finding
- sentinel/tests/generate_test_findings.py — generates synthetic findings to test all 4 rules
- sentinel/TEST_PLAN.md — full test plan and live results
- docs/sentinel-setup.md — end-to-end setup guide for free Azure trial

## Testing Evidence
- 12 findings ingested into OpenShieldFindings_CL confirmed
- Rule 1 tested: 7 HIGH/CRITICAL findings returned
- Rule 2 tested: Wave score 120 detected
- Rule 3 tested: 10 resources across 4 consecutive scans
- Rule 4 tested: Expected behaviour confirmed
- All 4 rules deployed and enabled in Microsoft Defender XDR

## Acceptance Criteria
- [x] Finding from scanner appears as data in Sentinel
- [x] KQL rules fire correctly on test data
- [x] Setup guide works end to end on free Azure trial

## Dependencies
Depends on #1 and #3 being merged to dev first.

PR #222, opened 2026-07-30
fix: resolve all open security alerts (8 Dependabot + 1 CodeQL)

TFT444:
## Summary

Resolves all 9 open security alerts tracked in #221: 8 Dependabot dependency vulnerabilities and 1 CodeQL code-scanning alert.

Closes #221.

## Dependency vulnerabilities (8 Dependabot alerts)

`react-router-dom` (root and `frontend/`) replaced with `react-router` 8.3.0. This is an import migration, not just a version bump - React Router v8 dropped the `react-router-dom` re-export package entirely. All 9 files that imported from `react-router-dom` now import from `react-router` directly:

- `frontend/src/App.jsx`
- `frontend/src/pages/AILayer.jsx`, `DetailedScan.jsx`
- `frontend/src/components/layout/Sidebar.jsx`, `Header.jsx`, `Layout.jsx`
- `frontend/src/components/scan/AskAIButton.jsx`
- `frontend/src/components/prioritization/QuickRemediation.jsx`
- `frontend/src/components/compliance/ComplianceTable.jsx`

Verified against the actual v8 package before making the change: `react-router/dom` only exports `RouterProvider`/`HydratedRouter` in v8 (nothing here uses either), while every export these files actually use (`BrowserRouter`, `Routes`, `Route`, `Navigate`, `useLocation`, `useNavigate`, `NavLink`, `Outlet`) lives in the main `react-router` package. Fixes alerts #12, #10, #8, #7, #6, #5.

`postcss` bumped to `>=8.5.18` (resolved to `8.5.25`). Fixes alert #9.

`brace-expansion` bumped via `npm audit fix` to `5.0.9`, not just the originally-requested `5.0.7` floor - `5.0.7` still carries a second, separate out-of-memory DoS advisory (`GHSA-mh99-v99m-4gvg`) that only clears at `5.0.8+`. Fixes alert #4 and that second advisory.

## CodeQL alert #31: clear-text logging of sensitive information

`scanner/rules/az_idn_006.py` logged the raw `endDateTime` value from a service principal's password credential when it failed to parse. Removed the value from the log line - `app_id` alone is sufficient to correlate the failure during debugging.

This matches the fix already proposed in draft PR #220 (Copilot Autofix) - this PR supersedes it, so #220 can be closed once this merges.

Also strengthened the existing regression test (`test_idn_006_malformed_end_date_time_does_not_log_key_id`) to assert the malformed `endDateTime` value itself never reaches the logs, not just `keyId`. Verified the test fails against the pre-fix code and passes after the fix.

## Other

Added `node_modules/` to the root `.gitignore`. The root `package.json` exists solely to give Dependabot a manifest to track `react-router` against (predates this PR), but regenerating its lockfile locally left an untracked `node_modules/` with nothing preventing an accidental commit.

## Test plan

- [x] `npm audit` clean at repo root (0 vulnerabilities)
- [x] `npm audit` clean in `frontend/` (0 vulnerabilities)
- [x] `npm run build` succeeds in `frontend/` (694 modules, no import errors)
- [x] `npm run lint` clean in `frontend/`
- [x] `ruff check .` and `ruff format --check` clean
- [x] Full `pytest -q`: 437 passed, 2 pre-existing/unrelated failures (local chromadb version mismatch, not present in CI)
- [x] Strengthened regression test verified to fail against pre-fix code, pass against the fix
- [x] `bandit` on the changed rule file: no issues

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -595,7 +595,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
-          node-version: "22"
+          node-version: "22.22.0"
           cache: npm
           cache-dependency-path: frontend/package-lock.json
 
--- a/.gitignore
+++ b/.gitignore
@@ -219,3 +219,6 @@ __marimo__/
 ai/vectorstore/
 .vercel
 .env*
+
+# Node (root package.json exists solely to track react-router for Dependabot)
+node_modules/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,13 +50,9 @@ from typing import Any, Dict, List
 
 RULE_ID = "AZ-STOR-001"
 RULE_NAME = "Public Blob Access Enabled on Storage Account"
-SEVERITY = "HIGH"           # HIGH / MEDIUM / LOW / INFO
-CATEGORY = "Storage"        # Storage / Network / Identity / Database / Compute / Key Vault / Kubernetes
-FRAMEWORKS = {
-    "CIS": "3.5",
-    "NIST": "PR.AC-3",
-    "ISO27001": "A.9.4.1"
-}
+SEVERITY = "HIGH"  # HIGH / MEDIUM / LOW / INFO
+CATEGORY = "Storage"  # Storage / Network / Identity / Database / Compute / Key Vault / Kubernetes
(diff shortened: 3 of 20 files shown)


ritiksah141 (changes requested):
Reviewed PR #222. One change is needed before approval:

React Router 8.3.0 requires Node >=22.22.0 and React/React DOM >=19.2.7. Please update the React dependency floors, declare the Node requirement, and ensure CI/Vercel uses a compatible version.

Verified the remaining security updates, audits, tests, build, lint, routing, and CI checks. Everything else looks sound.

Requesting changes for the runtime compatibility requirement.

ritiksah141 (approved):

PR #230, opened 2026-08-05
docs: update learn page statistics and add CI/CD workflow

### LINKED ISSUE #228: docs: update Learn page statistics and add CI/CD for auto-updates

`docs/learn/index.html` hardcodes repository statistics that were correct when the page was
written and have drifted since. The page currently advertises:

| Statistic | Page says | Actual |
|---|---|---|
| Azure scan rules | 39 | 66 |
| CLI remediation playbooks | 39 | 66 |
| High-severity checks | 22 | 38 |
| Hero terminal ("dynamic checks") | 39 | 66 |
| Severity card (HIGH / MEDIUM / LOW) | 22 / 13 / 4 | 38 / 23 / 4 |
| "Coverage by category" chart | omits Supply Chain and Kubernetes, Identity undercounted at 4 | 9 categories, Identity 15 |

`README.md` carries the same drift in four more places: the Misconfiguration Scanner and
Remediation Playbooks rows of the feature table, and both the `Scanner Engine` and
`Azure CLI Playbooks` nodes of the Mermaid architecture diagram (all reading 51).

Two problems follow from this:

1. **The Learn page understates the project.** New contributors and evaluators land on it and
   see roughly half the coverage OpenShield actually ships.
2. **It will drift again.** Every scanner-rule PR that merges to `dev` makes the numbers more
   wrong, and nothing in CI notices. Manual correction is not sustainable at the rate rules
   are being added.

## Describe the solution

Two parts:

**1. Correct the current numbers** in `docs/learn/index.html` — the headline counts, the
pipeline step, the section title, the intro paragraph, the hero terminal, the severity card
and the category chart all repeat the same figures and must be updated together. `README.md`
needs the same treatment for its four stale counts.

**2. Add a GitHub Actions workflow that keeps them correct.** On every push to `dev`:

- Derive the counts from the codebase rather than from a checked-in constant:
  - rule count from `scanner/rules/` (files that declare a `RULE_ID`)
  - playbook count from `playbooks/cli/*.sh`
  - severity counts from each rule's `SEVERITY` value
- Rewrite the statistics in `docs/learn/index.html` and `README.md`
- Commit and push only when something actually changed, tagged `[skip ci]` so the workflow
  cannot retrigger itself

The update script must be idempotent: running it on an already-current page produces no diff.

## Alternatives considered

- **Keep updating the page by hand.** Rejected — this issue exists precisely because that
  failed. The numbers went stale within a few release cycles.
- **A CI check that fails the build when the numbers are stale**, rather than auto-committing.
  Lower blast radius and no bot commits on `dev`, but it pushes the chore onto every
  contributor whose PR happens to change a count. Worth revisiting if the auto-commit
  approach turns out to conflict with branch protection on `dev`.
- **Render the statistics client-side** from a generated JSON file. Cleaner separation, but
  the Learn page is deliberately a single static HTML file with no build step, and adding one
  is out of proportion to the problem.

## Additional context

- Affected files: `docs/learn/index.html`, `README.md`
- New files: `.github/workflows/update-learn-page.yml`, `.github/scripts/update_learn_page.py`
- Implemented in PR #230
- No scanner, API, or frontend behaviour changes; documentation and CI only

parthrohit22:
## What does this PR do?

Closes out the remaining drift documented in #228. Fixes every stale count on the Learn page
(not just the five the first pass covered), fixes the four stale counts in `README.md`, and
extends `.github/scripts/update_learn_page.py` / `.github/workflows/update-learn-page.yml` so
both files are regenerated from the codebase on every push to `dev` and can no longer drift.

## Type of change
- [ ] New scan rule
- [ ] Remediation playbook
- [ ] Bug fix
- [ ] API endpoint
- [x] Documentation
- [ ] Compliance mapping

Not applicable — this PR adds no scanner rule. It only reads rule metadata to produce counts.

## Testing
- [ ] Tested against a real Azure free trial subscription — *not applicable, no Azure API calls*
- [ ] Returns correct JSON output — *not applicable, no scanner or API output*
- [x] All seven CI checks pass

Verification actually performed:

- `ruff check .` and `ruff format --check .` — clean repo-wide.
- `python -m bandit -r .github/scripts/` — no issues.
- `python -c "import yaml; yaml.safe_load(open('.github/workflows/update-learn-page.yml'))"` — valid.
- Derived every number from the filesystem before writing anything: `ls scanner/rules/az_*.py | wc -l` → 65,
  `ls playbooks/cli/*.sh | wc -l` → 65, `grep -l 'SEVERITY = "HIGH"' scanner/rules/az_*.py | wc -l` → 38.
  Cross-checked every rule's declared `PLAYBOOK` field against the filename-derived path — 0 orphans
  either direction.
- Idempotency, both files: reverted `docs/learn/index.html` and `README.md` to their pre-PR
  committed content, ran the script — output was byte-identical to the version already staged
  in this PR; a second run was a true no-op (`git diff` empty).
- Fail-loud path, tested in isolation: reworded the hero-terminal HTML so the regex could not
  match — the script correctly named exactly that pattern and exited 1, and confirmed no file
  was written on that failure.
- Empty-input guard, tested in isolation: pointed `RULES_DIR`/`PLAYBOOKS_DIR` at an empty temp
  directory — `main()` returned 1 and refused to write zeroes.
- Missing-`SEVERITY`/`CATEGORY` warning path, tested in isolation with synthetic rule files —
  correctly named the offending file without failing the run.
- Ran the actual DCO checker the repo uses in CI (`python scripts/check_dco.py <base> <head>`)
  against this branch — sign-off verified.

## Related issue

Closes #228

## Checklist
- [x] My code follows the rule template in CONTRIBUTING.md — *no new rule; Python follows the
      repo's Ruff/PEP 8 and type-hint standards*
- [ ] I added or updated the matching CLI playbook — *not applicable, no new rule*
- [ ] I added or updated all four compliance framework mappings — *not applicable, no new rule*
- [x] My branch name follows the convention — `docs/update-learn-page-and-add-cicd`, matching
      the `docs/description` form in CONTRIBUTING.md

## True numbers (derived from the filesystem, verified before any edit)

| Metric | Value |
|---|---|
| Rules (`scanner/rules/az_*.py` with `RULE_ID`) | 66 |
| Playbooks (`playbooks/cli/*.sh`) | 66 |
| Orphans (rule↔playbook) | 0 |
| HIGH / MEDIUM / LOW / CRITICAL severity | 38 / 23 / 4 / 1 |
| Categories | Identity 15, Network 15, Supply Chain 8, KeyVault 6, Kubernetes 6, Storage 5, Compute 4, Database 4, PostQuantum 3 |

Rebased onto `dev` after AZ-KV-006 merged in #235; the script picked the new rule up on its own
and moved KeyVault 5 → 6, MEDIUM 22 → 23 and both totals 65 → 66 with no hand-editing. That is
the drift-resistance this PR is for, so the numbers above will keep moving as rules land.

The 1 CRITICAL rule (`AZ-SC-005`) has no box in the Learn page's severity-distribution UI,
which is hardcoded to a 3-column HIGH/MEDIUM/LOW grid — documented as a known limitation, not
fixed here (see "Notes for reviewers").

## Changes

| File | Change |
|---|---|
| `docs/learn/index.html` | All previously-stale content fixed, not just the metric tiles: hero terminal `39 → 66 dynamic checks`; severity boxes HIGH `22 → 38`, MEDIUM `13 → 23` (LOW was already correct at 4); "Coverage by category" chart fully regenerated — previously omitted **Supply Chain (8)** and **Kubernetes (6)** entirely and undercounted Identity `4 → 15`; removed a "known cleanup item" note about `KeyVault` vs `Key Vault` naming that the chart fix resolves (the chart now renders the category string the code actually declares); rewrote the "Known gaps" card's "counts are checked and updated with each release" claim to accurately describe what's CI-generated vs. hand-maintained. |
| `README.md` | Fixed 4 stale counts: Misconfiguration Scanner feature-table row (`51 → 66`, and added the previously-unmentioned "supply chain" category), Remediation Playbooks row (`51 → 66`), and both Mermaid diagram nodes (`Scanner Engine`, `Azure CLI Playbooks`). |
| `.github/scripts/update_learn_page.py` | Extended to cover the hero terminal line, all three severity boxes (previously only HIGH), and to fully regenerate the category chart's rows/counts/bar-widths from real per-category data (a new category now appears automatically instead of being silently dropped). Added `render_readme()` so the same script now also fixes README.md. Every substitution is tracked via `re.subn`; if any pattern matches zero times the script prints which one and exits non-zero instead of silently doing nothing and exiting 0. Rule files with no parseable `SEVERITY`/`CATEGORY` are now warned about by name. Still idempotent. |
| `.github/workflows/update-learn-page.yml` | Extended to diff, `git add`, and commit `README.md` alongside `docs/learn/index.html`. Renamed to reflect the wider scope. SHA pins and `permissions: contents: write` unchanged from the first pass. |

## Decision: README.md is now automated, not left as a manual step

The task called for deciding — and stating clearly — whether to extend the workflow to also
keep `README.md` current, or leave that update as a one-time manual fix. I extended it: the
whole point of this issue is closing the exact class of drift where a doc's numbers silently
fall out of sync with the codebase, and `README.md`'s 4 stale counts are the same failure mode
in a second file. The substitution surface is small and well-bounded (2 feature-table rows, 2
Mermaid nodes), reuses the same tested `apply_replacements()`/fail-loud machinery already built
for the Learn page, and is the only choice consistent with the "Known gaps" card now stating
that these stats are CI-generated. `README.md`'s "30+ scan rules" roadmap checklist item and
other prose are unrelated to this PR's scope and were left untouched.

## Notes for reviewers

- Filed as a factual, no-marketing status comment on #228 before making any change, listing
  every stale location by file/line and the true numbers — see the issue thread.
- **Real discrepancy, not fixed in this PR (flagging for a follow-up issue):**
  `scanner/engine.py::load_rules()` — the code that actually executes rules at scan time —
  discovers rule modules with `RULES_DIR.glob("*.py")`, excluding only filenames starting with
  `_`. It does not require the `az_*.py` prefix or a `RULE_ID`, unlike `ci.yml`'s
  structure-validation job and this script, which both use the narrower convention. Today all
  three agree on 65 only because `scanner/rules/` happens to contain no stray file outside
  those two patterns — nothing enforces that. A leftover or misnamed `.py` file with a `scan()`
  function would be loaded and executed in production without being validated by CI or counted
  in either doc.
- **Branch protection on `dev`, from the first review round:** I don't have admin visibility
  into the repo's protection rules (only push access), so I can't fully confirm whether
  `GITHUB_TOKEN` can push directly to `dev`. `GET /repos/.../rules/branches/dev` (readable with
  plain read access) currently returns no effective rules, which suggests direct pushes are not
  blocked, but please verify in Settings → Branches before relying on that. If pushes are
  blocked, this workflow needs to open a PR instead of committing directly — I have not made
  that change silently.

### DIFF
--- a/.github/scripts/update_learn_page.py
+++ b/.github/scripts/update_learn_page.py
@@ -18,6 +18,7 @@ Running the script against already-current files produces no changes, which
 lets CI commit only on a real diff.
 """
 
+import html
 import re
 import sys
 from pathlib import Path
@@ -50,11 +51,39 @@ def count_rules() -> int:
     return sum(1 for path in RULES_DIR.glob(RULE_GLOB) if RULE_ID_PATTERN.search(path.read_text(encoding="utf-8")))
 
 
-def count_playbooks() -> int:
-    """Return the number of CLI remediation playbooks."""
-    if not PLAYBOOKS_DIR.is_dir():
-        return 0
-    return len(list(PLAYBOOKS_DIR.glob("*.sh")))
+def find_matching_playbooks() -> Tuple[int, List[str], List[str]]:
+    """Return (playbook_count, missing_for_rules, orphan_playbook_files).
+
+    Counting every *.sh file in playbooks/cli/ (the previous approach) also
+    picks up shared helper scripts - e.g. review_enterprise_resilience.sh,
+    which several fix_*.sh wrappers `exec` into rather than a playbook any
+    single rule owns - silently inflating the count past rule_count and
+    breaking the "every rule ships with a matching playbook" claim.
+
+    Instead this mirrors the naming convention ci.yml's playbook_check step
+    already enforces: playbooks/cli/fix_<rule filename stem>.sh for every
+    counted rule module. missing_for_rules lists rule files whose expected
+    playbook is absent; orphan_playbook_files lists *.sh files on disk that
+    no counted rule expects (informational - shared helpers are expected to
+    show up here and are not themselves an error).
+    """
+    if not RULES_DIR.is_dir() or not PLAYBOOKS_DIR.is_dir():
+        return 0, [], []
+
+    expected_names = set()
+    missing_for_rules: List[str] = []
+    for path in sorted(RULES_DIR.glob(RULE_GLOB)):
+        if not RULE_ID_PATTERN.search(path.read_text(encoding="utf-8")):
+            continue
+        expected_name = f"fix_{path.stem}.sh"
+        expected_names.add(expected_name)
+        if not (PLAYBOOKS_DIR / expected_name).is_file():
+            missing_for_rules.append(path.name)
+
+    actual_names = {p.name for p in PLAYBOOKS_DIR.glob("*.sh")}
+    orphan_playbook_files = sorted(actual_names - expected_names)
+    playbook_count = len(expected_names) - len(missing_for_rules)
+    return playbook_count, missing_for_rules, orphan_playbook_files
 
 
 def collect_rule_stats() -> Tuple[Dict[str, int], Dict[str, int], List[str], List[str]]:
@@ -64,6 +93,11 @@ def collect_rule_stats() -> Tuple[Dict[str, int], Dict[str, int], List[str], Lis
     files_missing_category). Severities and categories outside the values
     seen so far are still counted, keyed by whatever string the rule
     declares, so a new value is never silently dropped.
+
+    Files with no parseable RULE_ID are skipped entirely, matching
+    count_rules() - otherwise a non-rule file (excluded from rule_count)
+    could still be counted into these totals, making the severity boxes
+    disagree with the headline rule count.
     """
     severities: Dict[str, int] = {"CRITICAL": 0, "HIGH": 0, "MEDIUM": 0, "LOW": 0, "INFO": 0}
     categories: Dict[str, int] = {}
@@ -75,6 +109,8 @@ def collect_rule_stats() -> Tuple[Dict[str, int], Dict[str, int], List[str], Lis
 
     for path in sorted(RULES_DIR.glob(RULE_GLOB)):
         content = path.read_text(encoding="utf-8")
+        if not RULE_ID_PATTERN.search(content):
+            continue
 
         sev_match = SEVERITY_PATTERN.search(content)
         if sev_match:
@@ -109,7 +145,7 @@ def render_category_rows(categories: Dict[str, int]) -> str:
     for name, count in sorted(categories.items(), key=lambda item: (-item[1], item[0])):
         width = round(count / max_count * 100)
         rows.append(
-            f'            <div class="bar-row"><span>{name}</span>'
+            f'            <div class="bar-row"><span>{html.escape(name)}</span>'
             f'<div class="bar"><span style="width: {width}%;"></span></div>'
             f"<span>{count}</span></div>"
         )
@@ -224,7 +260,7 @@ def main() -> int:
             return 1
 
     rule_count = count_rules()
-    playbook_count = count_playbooks()
+    playbook_count, missing_playbooks, orphan_playbooks = find_matching_playbooks()
 
     if rule_count == 0 or playbook_count == 0:
         print(
@@ -232,6 +268,35 @@ def main() -> int:
             file=sys.stderr,
         )
         return 1
+
+    if missing_playbooks:
+        print(
+            f"Error: {len(missing_playbooks)} rule(s) have no matching playbook file: {', '.join(missing_playbooks)}",
+            file=sys.stderr,
+        )
+        return 1
+
+    # Belt-and-suspenders: find_matching_playbooks() only counts playbooks
+    # that resolve to a counted rule, so this should be unreachable once
+    # missing_playbooks is empty. Failing loudly here catches a future bug
+    # in that derivation rather than silently writing mismatched docs -
+    # this is exactly the drift class (rules and playbooks disagreeing)
+    # this script exists to catch.
+    if rule_count != playbook_count:
+        print(
+            f"Error: rule/playbook count mismatch (rules: {rule_count}, playbooks: "
+            f"{playbook_count}); refusing to write inconsistent docs.",
+            file=sys.stderr,
+        )
+        return 1
+
+    if orphan_playbooks:
+        print(
+            f"Warning: {len(orphan_playbooks)} playbook file(s) in playbooks/cli/ are not "
+            f"any rule's matching playbook and are excluded from the playbook count "
+            f"(expected for shared helpers other playbooks exec into): {', '.join(orphan_playbooks)}",
+            file=sys.stderr,
+        )
 
     severities, categories, missing_severity, missing_category = collect_rule_stats()
 
--- a/.github/scripts/update_learn_page.py
+++ b/.github/scripts/update_learn_page.py
@@ -248,6 +248,23 @@ def main() -> int:
             file=sys.stderr,
         )
 
+    chart_severities = {"HIGH", "MEDIUM", "LOW"}
+    excluded_severities = {
+        severity: count
+        for severity, count in severities.items()
+        if severity not in chart_severities and count
+    }
+    if excluded_severities:
+        excluded_detail = ", ".join(
+            f"{severity}: {count}" for severity, count in sorted(excluded_severities.items())
+        )
+        excluded_total = sum(excluded_severities.values())
+        print(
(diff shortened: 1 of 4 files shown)


ritiksah141 (dismissed):

ritiksah141 (changes requested):
**Blocking issue — `.github/workflows/update-learn-page.yml:47`**

`actions/checkout` leaves the runner in a detached `HEAD` state by default. As a result, the final bare `git push` has no current branch/upstream and will fail with `fatal: You are not currently on a branch`. The workflow therefore cannot publish the refreshed documentation after a push to `dev`.

Please either check out the branch explicitly or push the detached commit to the target branch, for example:

```yaml
git push origin HEAD:${{ github.ref_name }}
```

Please verify the commit-and-push path in CI before merging.

Vishnu2707:
@parthrohit22 , the conflicts and requested changes needs to be addressed to go ahead with the merge.

parthrohit22:
## Ready for review

PR #230 has been updated with the review follow-up:

- `.github/scripts/update_learn_page.py` now emits an explicit stderr warning when severity counts outside the HIGH/MEDIUM/LOW chart are excluded, including the excluded severity counts.
- The script was Ruff-formatted so the CI formatting check passes.

Validation completed:

- Ran the updater against `docs/learn/index.html` and `README.md`; it exited successfully, reported the expected `CRITICAL: 2` exclusion, and produced no documentation diff.
- Ruff check and format check pass.
- GitHub Actions CI is fully green, including CI Summary.

Review commits: `209b725`, `42ffda1`.

TFT444 (changes requested):
Great work @parthrohit22, the fail-loud substitution tracking and idempotent design are exactly right. One blocker:

1. **Rules/playbooks mismatch:** the regenerated docs now say 95 rules / 96 playbooks, but the README row still claims "Every rule ships with a matching playbook". Cause: `playbooks/cli/review_enterprise_resilience.sh` is an orphan no rule references. Your "0 orphans" check was true at 66/66 but dev moved. Please add a parity check to the script (fail/warn on rule count != playbook count), since this is exactly the drift class this PR exists to catch.

Two nits, same push if you like: `collect_rule_stats()` counts files without `RULE_ID` while `count_rules()` excludes them, so the severity boxes can disagree with the headline total; and `html.escape(name)` in `render_category_rows()` would be cheap insurance.

Fix the parity issue and this is good to go from my side.

parthrohit22:
@TFT444 Updated:

- **Parity check fixed.** The root cause was `count_playbooks()` counting every `*.sh` file in `playbooks/cli/`, which also picked up `review_enterprise_resilience.sh` -- a shared helper that several `fix_*.sh` wrappers `exec` into, not a playbook any single rule owns. Replaced it with `find_matching_playbooks()`, which mirrors the `fix_<rule filename stem>.sh` convention `ci.yml`'s `playbook_check` step already enforces, derives the count from actual rule-to-playbook matches, and fails loudly if a rule has no matching playbook or the two counts still disagree. Unmatched files are now a non-fatal warning instead of silently inflating the total. Regenerated docs now correctly read 95 rules / 95 playbooks, and the README's "every rule ships with a matching playbook" claim is true again.
- **Both nits fixed in the same push.** `collect_rule_stats()` now skips files with no parseable `RULE_ID`, matching `count_rules()`'s own filter, so the severity-box totals can't disagree with the headline count. `render_category_rows()` now HTML-escapes the category name before embedding it.

Ran the updater against the current tree -- clean exit, no further diff. Ruff check/format clean. CI is green including CI Summary.

Latest commit: `780113a`.

ritiksah141 (approved):

ritiksah141:
@parthrohit22, all good just need to be rebased once the #317 Lands with the dev and good to go

parthrohit22:
@ritiksah141 Rebased onto the latest dev (9a84012). Ran the generator afterward - no drift, statistics are already current. Full suite and ruff still clean.

One CI note so it's not mistaken for something new: `SCA (pip-audit)` is currently failing on this branch because `dev` still has `chromadb==0.4.24` (CVE-2026-45830/45833) - that's #317, not anything in this PR. It's approved and just hasn't merged yet. Once it lands I'll rebase again and this clears on its own, as you flagged.

TFT444 (approved):
The requested fixes are addressed and the current checks pass. Approving.

m-khan-97:
@parthrohit22, this is approved and green, but it is now two commits behind `dev`. Please rebase once more and rerun CI so the generated statistics are verified against the actual merge base.

PR #208, opened 2026-07-16
governance: enforce DCO and document continuity checks

TFT444:
## Summary

Implements the repository-controlled parts of #205.

- adopts DCO 1.1 as the contribution authorization mechanism
- adds a dependency-free PR check requiring a valid `Signed-off-by` trailer on every introduced commit
- documents contributor sign-off and repair instructions
- adds the DCO requirement to the pull request checklist
- adds an owner-verifiable continuity capability matrix and six-month review process
- adds regression tests for valid, missing, malformed, and mixed commit sign-offs

## Validation

- 3 DCO helper tests pass
- this PR commit passes the new DCO checker
- Ruff and formatting pass
- workflow YAML parses successfully
- `git diff --check` passes

## Owner confirmation still required

Issue #205 must remain open until an organization owner privately verifies that at least two people can administer issues, merge approved work, publish/recover releases, and recover critical organization/deployment access. Documentation cannot prove that underlying access state.

Advances #205

PR #232, opened 2026-08-05
fix(ci): apply DCO enforcement baseline

TFT444:
## Summary

Fixes false DCO failures caused by historical commits predating enforcement, especially on squash-based dev to main release pull requests.

- subtracts the ancestry of the explicit PR #208 DCO baseline from the PR commit range
- continues checking new work from branches created before the baseline
- preserves merge-commit exclusion
- treats an all-exempt range as successful
- documents the cutoff and adds real Git-history regression coverage

## Validation

- 6 passed in 	ests/test_dco_check.py
- Ruff lint and format checks pass
- DCO workflow YAML parses successfully
- replay of PR #227 base/head SHAs passes and verifies the one post-policy commit

Closes #231

### DIFF
--- a/.github/workflows/dco.yml
+++ b/.github/workflows/dco.yml
@@ -21,4 +21,7 @@ jobs:
         env:
           BASE_SHA: ${{ github.event.pull_request.base.sha }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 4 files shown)


Vishnu2707 (approved):
Approved.

PR #233, opened 2026-08-06
fix(security): complete allowlist input validation audit

### LINKED ISSUE #201: security: complete allowlist-based input validation audit
## Goal

Audit every untrusted input boundary and ensure restricted values are validated with explicit allowlists, types, formats, and size limits. This supports the OpenSSF Silver `input_validation` criterion tracked in #199.

## Scope

Review inputs entering through:

- API JSON bodies, path values, and query parameters
- Scan, finding, subscription, resource, and framework identifiers
- AI questions and provider settings
- Playbook and file-path selection
- Sentinel ingestion data
- Website editor fields and external URLs

## Requirements

- Centralize reusable validation where practical.
- Validate allowed values, syntax, type, and length before database, file, Azure, subprocess, or external-service use.
- Reject invalid input with consistent `400` responses without exposing internal details.
- Do not rely only on denylists of known malicious strings.
- Preserve legitimate API behavior and document intentional unrestricted fields.
- Add tests for valid, invalid, missing, oversized, malformed, and injection-style inputs.

## Completion

- [ ] Every input boundary is recorded and reviewed
- [ ] Identified validation gaps are fixed
- [ ] Security regression tests cover all changed boundaries
- [ ] API documentation describes relevant limits
- [ ] Ruff, frontend checks, full tests, CodeQL, Bandit, and Semgrep pass
- [ ] The Silver evidence register links to the completed audit

Implementation should begin only after the boundary inventory confirms the exact scope.

TFT444:
## Summary
- add centralized allowlist, type, format, and size validation for API inputs
- validate scan/findings paths, query parameters, AI payloads, request IDs, auth headers, and Sentinel ingestion data
- document every reviewed boundary and update the OpenSSF Silver evidence register
- add security regression coverage for malformed, oversized, duplicated, and injection-style inputs

## Validation
- `ruff check .`
- `ruff format --check .`
- `511 passed, 3 skipped` with `84.42%` API/scanner coverage
- frontend lint, i18n, accessibility, and production build
- website URL security tests
- Bandit: no medium/high findings
- pip-audit and both npm audits: 0 vulnerabilities

Closes #201

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -29,6 +29,7 @@ _MAX_CONTENT_LENGTH = 2 * 1024 * 1024  # 2 MB
 
 _INSECURE_JWT_DEFAULT = "change-me-in-production"
 _MIN_JWT_SECRET_LENGTH = 32
+_MAX_AUTHORIZATION_HEADER_LENGTH = 8192
 _GENERATE_CMD = 'python -c "import secrets; print(secrets.token_urlsafe(32))"'
 
 
@@ -167,7 +168,7 @@ def create_app() -> Flask:
             return None
 
         auth = request.headers.get("Authorization", "")
-        if not auth.startswith("Bearer "):
+        if len(auth) > _MAX_AUTHORIZATION_HEADER_LENGTH or not auth.startswith("Bearer "):
             return jsonify(
                 {
                     "error": "Missing or malformed Authorization header",
--- a/api/observability.py
+++ b/api/observability.py
@@ -13,6 +13,7 @@ status code, scan status, rule ID and provider name) are used as labels.
 
 import logging
 import os
+import re
 import time
 import uuid
 
@@ -35,6 +36,7 @@ except ImportError:  # pragma: no cover
 logger = logging.getLogger(__name__)
 
 REQUEST_ID_HEADER = "X-Request-ID"
+_REQUEST_ID_RE = re.compile(r"^[A-Za-z0-9._:-]{1,128}$")
 
 # --------------------------------------------------------------------------- #
 # Prometheus metrics                                                           #
@@ -160,7 +162,8 @@ def init_app(app: Flask) -> None:
 
     @app.before_request
     def _start_observability() -> None:
-        g.request_id = request.headers.get(REQUEST_ID_HEADER) or str(uuid.uuid4())
+        supplied_request_id = request.headers.get(REQUEST_ID_HEADER, "")
+        g.request_id = supplied_request_id if _REQUEST_ID_RE.fullmatch(supplied_request_id) else str(uuid.uuid4())
         g.request_start_time = time.perf_counter()
 
     @app.after_request
--- a/api/routes/ai.py
+++ b/api/routes/ai.py
@@ -8,6 +8,18 @@ from flask import Blueprint, jsonify, request
 from api.rate_limit import rate_limit
 from api.services.ai_provider import PROVIDERS as SUPPORTED_PROVIDERS
 from api.services.ai_provider import get_completion
+from api.validation import (
+    MAX_API_KEY_LENGTH,
+    MAX_MODEL_LENGTH,
+    MAX_QUESTION_LENGTH,
+    MODEL_RE,
+    ValidationError,
+    bounded_string,
+    choice,
+    findings_list,
+    reject_unknown_fields,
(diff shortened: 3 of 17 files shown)


parthrohit22 (changes requested):
overall nice work on this @TFT444  the validator module is clean and I like that you're returning a
generic error instead of leaking the exception text. Pulled it down and ran it: your tests
pass, full suite and ruff are fine on my end.

One thing blocking though. `?category=` is now case-sensitive, but the query underneath
still does `LOWER(category) = LOWER(%s)`, so anything lowercase that worked before now
400s — `?category=network`, `?category=storage`, `?category=keyvault` all break. You
handled this on severity with `case="upper"`, category just missed it. Can't use `.title()`
since it'd mangle "Key Vault", so probably a lowercase-keyed lookup back to the canonical
value. Worth a test pinning it too.

Couple of smaller things while you're in there:

- `/api/compliance/<framework>` still echoes the raw input back in its error and doesn't
  go through the new validators — feels like it should, given it's in scope.
- `validate_config()` in the Sentinel CLI isn't wrapped, so a missing env var dumps a
  traceback now. Also one bad record kills the whole batch, which is fine as a decision,
  just worth documenting.

Rest looks good. Fix the category thing and I'm happy with it.


Reference :  here's what I checked on my end: pulled the branch and ran your two new
test files (28 passed) plus the full suite — 506 passed, 3 skipped. The 5 failures I hit
are already there on `dev` from missing optional deps locally, nothing to do with your
changes. `ruff check` and `ruff format --check` both clean. Also poked at the routes
manually with some junk input — SQL-ish strings, script tags, non-canonical UUIDs — and
they all 400 before touching the DB, with nothing reflected back. Checked the category
allowlist against every `CATEGORY` constant in `scanner/rules/` too, all 9 are covered.
That's how the case-sensitivity thing turned up — compared behaviour side by side
against `dev`.

TFT444:
> overall nice work on this @TFT444 the validator module is clean and I like that you're returning a generic error instead of leaking the exception text. Pulled it down and ran it: your tests pass, full suite and ruff are fine on my end.
> 
> One thing blocking though. `?category=` is now case-sensitive, but the query underneath still does `LOWER(category) = LOWER(%s)`, so anything lowercase that worked before now 400s — `?category=network`, `?category=storage`, `?category=keyvault` all break. You handled this on severity with `case="upper"`, category just missed it. Can't use `.title()` since it'd mangle "Key Vault", so probably a lowercase-keyed lookup back to the canonical value. Worth a test pinning it too.
> 
> Couple of smaller things while you're in there:
> 
> * `/api/compliance/<framework>` still echoes the raw input back in its error and doesn't
>   go through the new validators — feels like it should, given it's in scope.
> * `validate_config()` in the Sentinel CLI isn't wrapped, so a missing env var dumps a
>   traceback now. Also one bad record kills the whole batch, which is fine as a decision,
>   just worth documenting.
> 
> Rest looks good. Fix the category thing and I'm happy with it.
> 
> Reference : here's what I checked on my end: pulled the branch and ran your two new test files (28 passed) plus the full suite — 506 passed, 3 skipped. The 5 failures I hit are already there on `dev` from missing optional deps locally, nothing to do with your changes. `ruff check` and `ruff format --check` both clean. Also poked at the routes manually with some junk input — SQL-ish strings, script tags, non-canonical UUIDs — and they all 400 before touching the DB, with nothing reflected back. Checked the category allowlist against every `CATEGORY` constant in `scanner/rules/` too, all 9 are covered. That's how the case-sensitivity thing turned up — compared behaviour side by side against `dev`.

changes have been doen can u re-check the pr ???if everything ok please approve it.
thank you

parthrohit22 (approved):
nice work @TFT444 ,
This is the one I spent the most time on given it touches auth and every route. Read through the whole diff, not just the test output. The UUID validator specifically stood out — you're guarding against the "pass a 32-char no-hyphen UUID and hope it slips past a naive check" bypass by comparing against the canonical str(parsed) form, which a lot of validation-audit PRs miss. The auth-header length cap and the request-ID regex before it hits the logs are both good defensive adds. Ran the full suite — 589 passed (33 new), zero regressions, ruff clean. This closes out #201 which has been sitting since mid-July. Nothing to flag, approving as-is.

PR #237, opened 2026-08-07
fix: resolve real Disk resource for AZ-CMP-002 encryption detection

parthrohit22:
## What does this PR do?
Fixes AZ-CMP-002, which was fail-open and could never produce a finding against real Azure data under any configuration.

## Type of change
- [ ] New scan rule
- [ ] Remediation playbook
- [x] Bug fix
- [ ] API endpoint
- [ ] Documentation
- [ ] Compliance mapping

- Rule ID: AZ-CMP-002
- Severity: HIGH (confirmed violation) / LOW (indeterminate — see "Review changes" below)
- Category: Compute
- Frameworks mapped: CIS 7.2 / NIST PR.DS-1 / ISO 27001 A.10.1.1 / SOC 2 CC6.7 (mapping unchanged — only detection logic was broken)

## Testing
- [x] Tested against a real Azure free trial subscription
- [x] All seven CI checks pass

**Real Azure test:** I have an "Azure for Students" subscription authenticated via `az login`. I provisioned a throwaway VM (`Standard_B2ats_v2`, no public IP) in a disposable resource group — its default OS disk is `EncryptionAtRestWithPlatformKey` with no ADE, i.e. a genuinely non-compliant disk. I ran the rule against it live via `AzureCliCredential`:
- **Old (pre-fix) implementation** against that real VM → **0 findings** (confirms the fail-open bug against real data, not just theory).
- **Fixed implementation** against the same real VM → **1 finding**, `determination: "non_compliant"`, correct disk name, correct framework mapping.

The resource group was deleted immediately after (`az group delete --yes`) — confirmed fully gone before pushing this branch, no lingering billable resources.

I also ran `pytest`, `ruff check`, `ruff format --check`, and `bandit -r api/ scanner/ ai/ -ll` locally exactly as the CI workflow configures them — 490 passed, 3 skipped. The actual GitHub Actions run on this push has now completed: every required check (lint, rule/compliance validation, secret scan, SAST, SCA, SBOM, container scan, backend tests + coverage, frontend, website, DCO sign-off, `CI Summary`) is green. The only non-green item is `semgrep-cloud-platform/scan`, an external app-based check outside `ci.yml`'s own required job set, still `pending` as of this update.

## Related issue
Closes #236

## Checklist
- [ ] I added or updated all four compliance framework mappings
- [x] My branch name follows the convention: `fix/description` (bug fix, per CONTRIBUTING.md)

The `FRAMEWORKS` dict and `compliance/frameworks/*.json` entries for AZ-CMP-002 were already correct and are unchanged — the bug was purely that the rule's detection logic could never fire, so every one of those mappings was a guaranteed silent pass. No mapping file changes were needed.

---

### The defect

`_disk_needs_flagging()` in `scanner/rules/az_cmp_002.py` read `managed_disk.security_profile` and `managed_disk.encryption`, then `.type` off whichever was non-`None`. The object actually passed in is `azure.mgmt.compute.models.ManagedDiskParameters`, which only exposes `id`, `storage_account_type`, `disk_encryption_set`, and `security_profile` — there is no `encryption` attribute. Its `security_profile` is a `VMDiskSecurityProfile`, which carries `security_encryption_type`, not `type`. Every branch resolved to `None`, so the function returned `False` for every disk on every VM, unconditionally.

`tests/test_rules_compute.py` didn't catch it because the fixtures built `managed_disk=make_resource(encryption=make_resource(type=...))` — a shape the SDK never returns, since `make_resource` accepts arbitrary kwargs.

### The fix

- Added `AzureClient.get_disk()` to resolve the real `Disk` resource from a managed disk's `id`, cached per subscription, returning `None` on failure (matching the existing convention on `get_managed_clusters`/`get_container_registries` — callers must never treat `None` as compliant).
- `_classify_disk()` now determines compliance from the real `Disk` resource: `Disk.encryption.type` distinguishes `EncryptionAtRestWithPlatformKey` (non-compliant) from `EncryptionAtRestWithCustomerKey`/`EncryptionAtRestWithPlatformAndCustomerKeys` (compliant), and `Disk.encryption_settings_collection.enabled` reports Azure Disk Encryption (compliant regardless of key type).
- Disks whose `Disk` resource can't be read are no longer silently treated as compliant — they're logged and surfaced so this can't regress into a silent pass again.
- Updated `playbooks/cli/fix_az_cmp_002.sh` with a note that it only remediates confirmed (`non_compliant`) findings, not indeterminate ones, since running ADE remediation against an indeterminate finding could unnecessarily touch an already-compliant CMK disk.
- Rebuilt the AZ-CMP-002 fixtures against the real SDK attribute surface, covering compliant (customer key, platform-and-customer key, ADE enabled), non-compliant (platform key only), and indeterminate (unreadable `Disk`) cases. The non-compliant test is a regression test — it fails against the old implementation and passes against this fix.

### Review changes (per @TFT444)

- **Indeterminate disks now report a distinct `LOW`-severity "unknown scan result"**, not a `HIGH` "confirmed" finding — separate `INDETERMINATE_SEVERITY`/`INDETERMINATE_DESCRIPTION`/`INDETERMINATE_REMEDIATION`. A confirmed non-compliant disk still keeps the finding at `HIGH` even if a sibling disk on the same VM is indeterminate (`metadata.determination` distinguishes the two; see `test_cmp_002_confirmed_violation_outweighs_indeterminate_sibling_disk`). This still satisfies "never silently pass on an unreadable disk," just without inflating an unknown into a confirmed HIGH violation.
- **Added regression tests built from genuine `azure.mgmt.compute.models` instances** (`ManagedDiskParameters`, `Disk`, `Encryption`, `EncryptionSettingsCollection`) rather than the `make_resource` stand-in, which accepts arbitrary kwargs and is exactly what let the original bug's fixtures assert against a shape the SDK never returns. One test (`test_cmp_002_managed_disk_parameters_has_no_encryption_attribute`) asserts directly that `ManagedDiskParameters` has no `encryption` attribute, as a canary against this exact bug class recurring if the SDK's shape ever drifts. These follow the same `@pytest.mark.skipif(not _AZURE_SDK_AVAILABLE, ...)` pattern already used in `tests/test_rules_network.py` for AZ-NET-003.

### DIFF
--- a/playbooks/cli/fix_az_cmp_002.sh
+++ b/playbooks/cli/fix_az_cmp_002.sh
@@ -3,6 +3,16 @@
 # Rule: AZ-CMP-002 — Virtual machine disk not protected by CMK or ADE
 # Usage: ./fix_az_cmp_002.sh <resource-group> <vm-name> <keyvault-name>
 # Severity: HIGH
+#
+# This script only remediates a *confirmed* finding (metadata.determination
+# == "non_compliant" — a disk resolved to Disk.encryption.type ==
+# EncryptionAtRestWithPlatformKey with no ADE enabled). If the finding's
+# metadata.determination is "indeterminate", the scanning principal could
+# not read the Disk resource (missing Microsoft.Compute/disks/read, or the
+# disk was deleted) and the actual encryption state is unknown. Running this
+# script against an indeterminate finding may enable ADE on a disk that was
+# already compliant via CMK. Run `az disk show --ids <disk-id>` to confirm
+# the encryption state before proceeding in that case.
 
 set -e
 
@@ -17,6 +27,8 @@ if [ -z "$RESOURCE_GROUP" ] || [ -z "$VM_NAME" ] || [ -z "$KEYVAULT_NAME" ]; the
   echo "  1. Create a Key Vault if one does not exist:"
   echo "     az keyvault create --resource-group <rg> --name <kv-name> --enabled-for-disk-encryption true"
   echo "  2. Ensure the VM is running before enabling encryption"
+  echo "  3. If the finding was indeterminate, confirm the actual encryption"
+  echo "     state with 'az disk show' before running this script."
   exit 1
 fi
 
--- a/scanner/azure_client.py
+++ b/scanner/azure_client.py
@@ -55,6 +55,7 @@ class AzureClient:
         self._managed_identity_principals_cache: Any = _UNSET
         self._subscription_role_assignments_cache: Any = _UNSET
         self._container_registries_cache: Any = _UNSET
+        self._disks_cache: Dict[str, Any] = {}
         self.devops_client = self._build_devops_client()
 
     def _build_devops_client(self) -> Optional[Any]:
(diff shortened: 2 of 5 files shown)


TFT444 (commented):
@parthrohit22  The main fix looks good and all checks passed. Please handle unreadable disks as an unknown scan result instead of a confirmed HIGH finding, and add one test using the real Azure SDK disk model so the original mock-related bug cannot happen again.

ritiksah141 (approved):

TFT444 (approved):
The fix that was required is done.

PR #242, opened 2026-08-08
chore(scanner): add placeholder for storage Account HTTPS enforcement…

Ablaze005:
## Summary

Add placeholder rule and remediation for **AZ-STOR-006 — Storage Account HTTPS-only enforcement disabled**.

This PR introduces:
- `openshield/rules/storage/az_stor_006.py` — rule that detects storage accounts where `supportsHttpsTrafficOnly` is false.
- `playbooks/cli/fix_az_stor_006.sh` — remediation playbook to enable HTTPS-only via Azure CLI.
- `openshield/tests/storage/test_az_stor_006.py` — minimal unit test skeleton.

## Why

Reviewers requested a storage HTTPS enforcement rule placeholder that follows the repository rule pattern and includes a remediation playbook. This PR mirrors the style and scan pattern used by existing storage rules (for example, AZ-STOR-004) and provides a minimal, testable remediation script.

## How to test

1. Ensure Azure CLI is installed and you are logged in with appropriate permissions.
2. Run unit tests:
image

## Notes for maintainers

- The rule expects the `azure_client` to implement:
- `get_storage_accounts()`
- `parse_resource_id(resource_id)`
- `get_storage_account_https_only(resource_group, account_name)` returning `True|False|None`
If your `AzureClient` uses different method names, update the rule or let me know and I will adapt it.

- The playbook should be executable in the commit. If it is not, run:

chmod +x playbooks/cli/fix_az_stor_006.sh

and push the change.

## Linked issue

Related to: #<245>

## Commit and branch

- Branch: `feat/az-stor-006`
- Commit message: `feat(storage): add AZ-STOR-006 HTTPS-only enforcement rule and playbook`


parthrohit22 (changes requested):
Sorry, this one needs to go back — and honestly I don't think it should come back as AZ-STOR-006 at all. AZ-STOR-002 already does this exact check (HTTPS-only enforcement on storage accounts), same severity, same category. If this merges we'd be reporting the same misconfiguration twice under two different rule IDs, which is the opposite of what we're trying to fix in [#263](https://github.com/openshield-org/openshield/issues/263) right now.

On top of that, the rule doesn't actually work — I merged it and ran the tests, and its own test (test_stor_006_noncompliant_returns_one_finding) fails. It's reading supportsHttpsTrafficOnly off the account object/properties, but that's the ARM REST API field name, not the Python SDK attribute — the real one is enable_https_traffic_only (you can see it used correctly in AZ-STOR-002 already). So right now this rule would silently report zero findings against a real subscription.

@Ablaze005 I know you've already been through a few rounds on this with @ritiksah141 — I'd say don't keep patching this one, close it out and pick up an uncovered control instead, e.g. [#246](https://github.com/openshield-org/openshield/issues/246) (SQL TLS enforcement). HTTPS-only is already covered and working.

Ablaze005:
Thanks for the advice. I will work on issue 246 if it is assigned to me

Vishnu2707:
@Ablaze005, CI seems to fail - please cross check and fix the issues, also do have a look at the changes requested.

TFT444 (changes requested):
Thanks for the contribution, but I have to recommend closing (or fully repurposing) this PR rather than iterating — there are three independent disqualifiers:

**1. AZ-STOR-006 duplicates an existing rule.**
`scanner/rules/az_stor_002.py` ("Storage Account Allows HTTP Traffic (Not HTTPS-Only)", HIGH, CIS 3.1) already checks `enable_https_traffic_only` on the same resource type. This PR re-implements the same control under a new ID, which would double-report every finding.

**2. The rule can never fire against real Azure objects.**
The fallback chain checks a `get_storage_account_https_only()` client method (exists on neither `AzureClient` nor `MockAzureClient`), then `account.properties.get(...)` (SDK properties are flattened attributes, not a dict), then `getattr(account, "supportsHttpsTrafficOnly", None)` — but the Python SDK attribute is `enable_https_traffic_only`. The result is always `None`, so every account is skipped. The failing Backend Tests check confirms this.

**3. Mappings are wrong/missing.**
CIS 3.2 is infrastructure encryption — secure transfer is 3.1, which AZ-STOR-002 already claims. And no entries were added to the four `compliance/frameworks/*.json` files, which Rule & Compliance Validation requires.

Also: DCO sign-off, ruff, Trivy, and CI Summary are red (5 failing checks total).

If you'd like to keep contributing a storage rule, I'd suggest picking a control that isn't covered yet (e.g. from the scanning-gaps list in #259 or the storage items in the roadmap) and we can help scope it.

PR #250, opened 2026-08-11
feat(scanner): add Azure Resource Graph inventory foundation

### LINKED ISSUE #249: feat(scanner): introduce high-speed Azure resource scanning
## Problem

OpenShield currently discovers Azure resources through separate SDK calls. This works for individual rules, but repeated service calls can become slow across large subscriptions and do not provide one reusable inventory for attack-path correlation.

This issue introduces Azure Resource Graph (ARG) as the first foundation of the wider OpenShield Evidence Graph architecture.

## Proposed work

- Add an Azure Resource Graph client using the existing Azure credential flow.
- Query resources across authorised subscriptions in batches.
- Handle pagination, throttling, permissions and partial failures safely.
- Convert ARG results into a normalised OpenShield resource snapshot.
- Preserve tenant and subscription boundaries in every resource record.
- Record collection time, resource count and query errors.
- Keep targeted Azure SDK calls for security properties unavailable through ARG.
- Benchmark ARG discovery against the current SDK-based approach.
- Add offline unit tests without requiring a live Azure subscription.

## First proof of concept

Collect a defined set of Azure resources through ARG, create one normalised snapshot and compare its speed and completeness with the current scanner.

## Acceptance criteria

- ARG inventory works across one or more authorised subscriptions.
- Every record contains resource ID, type, name, location, subscription and resource group.
- Pagination and throttling are tested.
- Empty, partial and failed queries remain distinguishable.
- Results from different tenants or scans cannot be mixed.
- Existing scanner behaviour remains available as a fallback.
- No production resource is modified.
- Performance results are documented as measurements, not guarantees.
- Ruff, tests and security CI checks pass.

## Non-goals

- Replacing every Azure SDK call in this issue.
- Building the complete attack graph.
- Automatic remediation.
- Guaranteeing sub-30-second scans before benchmarking.

## Future use

The normalised ARG snapshot will later support attack-path correlation, Code-to-Cloud tracing and verified remediation.

TFT444:
## Summary

Introduces Azure Resource Graph inventory as Phase 1 of the proposed OpenShield Evidence Graph.

- Adds paginated multi-subscription ARG collection.
- Produces normalised, tenant-scoped resource snapshots.
- Preserves complete, partial, and failed collection states.
- Handles throttling with bounded retries and detects repeated pagination tokens.
- Records snapshot identity, collection duration, page count, resources, and safe errors.
- Adds the Evidence Graph and Customer Trust Layer architecture document.
- Keeps existing SDK rule execution unchanged while ARG completeness is evaluated.

## Security boundaries

- ARG collection is read-only.
- Every resource is bound to a tenant, authorised subscription, and snapshot.
- Cross-tenant and unauthorised-subscription rows are rejected.
- Phase 1 does not persist snapshots or send them to an LLM or external service.
- No remediation or production write access is introduced.

## Verification

- ARG regression suite: 10 passed, 92% module coverage.
- Broader scanner/API suite: 536 passed, 3 skipped.
- Ruff check and format pass.
- Python compilation passes.
- Bandit reports no medium or high findings.
- Installed-environment dependency audit reports no known vulnerabilities.
- Detects `resultTruncated` responses without a continuation token and marks the snapshot partial.
- GitHub CI: all 20 checks pass on the final fix revision.

The unrelated ChromaDB dependency test was excluded from the local Windows run because that dependency could not be installed cleanly in the isolated environment. GitHub CI ran the repository's complete Linux suite successfully.

## Live Azure validation

Validated on 11 August 2026 in Azure Resource Graph Explorer using an active Azure for Students subscription containing five resources. The scope was restricted to that student subscription; two unrelated university production subscriptions were not selected.

The OpenShield inventory projection was executed as a read-only query:

```kusto
Resources
Resources
| project id, name, type, location, subscriptionId, resourceGroup, tenantId, tags, properties
| summarize resourceCount = count()
| extend validation = "OpenShield PR #250 ARG inventory"
```

Results:

- Aggregate resource count: 5 resources in 0.442 seconds.
- Full OpenShield inventory projection: 5 of 5 resources in 0.575 seconds.
- No Azure resources, configuration, or permissions were changed.
- A privacy-safe proof query labelled the result `OpenShield PR #250 ARG inventory` without exposing resource identifiers in shared evidence.

This proves the KQL projection and real Azure ARG execution path. It does not yet prove the local Python client against live university credentials because the local Azure CLI remains authenticated to a different tenant.

## Remaining benchmark gates

- Run `ArgInventoryClient` directly with authorised live Azure CLI credentials.
- Exercise pagination with more than one result page.
- Compare ARG completeness with the current SDK scanner.
- Record properties that require targeted SDK enrichment.
- Validate throttling, partial-scope, and restricted-permission behaviour.

No sub-30-second performance guarantee is claimed until these larger-scale gates are complete.

Closes #249

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ OpenShield uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- Azure Network Layer Assurance API with 20-domain coverage, network-rule classification, and authoritative IP forwarding and direct Internet route checks
 - Azure Resource Graph inventory snapshots as the first OpenShield Evidence Graph foundation
 - Azure Data Link Layer Assurance API with LLC and MAC coverage plus ExpressRoute Direct MACsec checks
 - Azure public-cloud Physical Layer Assurance API with complete OSI and IEEE PHY domain, sublayer, and provider-evidence coverage
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ OpenShield uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Added
 
+- Azure Resource Graph inventory snapshots as the first OpenShield Evidence Graph foundation
 - Azure Data Link Layer Assurance API with LLC and MAC coverage plus ExpressRoute Direct MACsec checks
 - Azure public-cloud Physical Layer Assurance API with complete OSI and IEEE PHY domain, sublayer, and provider-evidence coverage
 - Semgrep SAST integrated into GitHub Actions CI as an open-source, account-free complement to CodeQL
--- a/docs/openshield-evidence-graph.md
+++ b/docs/openshield-evidence-graph.md
@@ -11,21 +11,115 @@ OpenShield currently evaluates many security controls independently. The propose
 resources, identities, permissions, exposures, findings, infrastructure code, and remediation evidence into one
 traceable security model.
 
-```text
-Azure Resource Graph and targeted SDK evidence
-                    |
-                    v
-          Normalised evidence snapshot
-                    |
-                    v
-          Evidence-backed attack paths
-                    |
-                    v
-        Reviewed, controlled remediation
-                    |
-                    v
-             Rescan and verification
+## Target architecture
+
+The target is a hybrid scanner. ARG performs broad resource discovery, Python remains the orchestration and rule
+engine, and Azure SDK calls are retained only where ARG cannot supply the required evidence.
+
+```mermaid
+flowchart TB
+    subgraph Scope[Customer Azure boundary]
+        A[Azure tenant]
+        B[Authorised subscriptions]
+        C[Least-privilege read credential]
+        A --> B --> C
+    end
+
+    subgraph Collection[Phase 1 - inventory foundation]
+        D[ARG KQL batch query]
+        E[Pagination and bounded retry]
+        F[Truncation and repeated-token guards]
+        G[Normalised inventory snapshot]
+        H{Snapshot status}
+        I[COMPLETE]
+        J[PARTIAL]
+        K[FAILED]
+        C --> D --> E --> F --> G --> H
+        H --> I
+        H --> J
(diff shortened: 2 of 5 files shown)


ritiksah141 (dismissed):

ritiksah141 (approved):

m-khan-97 (approved):
Reviewed and fetched the branch to run the tests directly (installed azure-mgmt-resourcegraph==8.0.1 locally to confirm the pin is real and importable): 11/11 pass.

This is well-built, self-contained foundational work — it's a new standalone module (`scanner/arg_inventory.py`) not yet wired into `ScanEngine`, so there's zero risk to the existing 74-rule scan path. What stood out:

- **Tenant/subscription isolation is actually enforced in code**, not just asserted in the docs — `_normalise_resource` rejects any row whose `subscriptionId` isn't in the authorised set or whose `tenantId` doesn't match, and it's tested (`test_rejects_cross_tenant_and_unauthorised_subscription_rows`). Given the doc's "Customer Trust Layer" section makes tenant isolation a headline guarantee, I specifically checked this claim holds in the implementation — it does.
- **Failure semantics are correct and well-tested**: COMPLETE only with zero errors, PARTIAL preserves already-collected resources when a later page fails, FAILED when nothing was ever collected. Matches the documented contract exactly, including the repeated-pagination-token guard (infinite-loop protection) and the truncated-without-token case.
- **`_normalise_subscriptions` guards against the classic `isinstance(x, str)` foot-gun** (passing a bare string and having it iterated character-by-character) — small thing, but the kind of bug that's easy to miss and annoying to debug.
- Retry-After handling is sensible (parses the header, falls back to capped exponential backoff if it's not a plain number), and retries are correctly scoped to 429/503 only.

The architecture doc is honest about scope — explicitly flags the benchmark numbers as not yet an apples-to-apples comparison until rules actually migrate to the shared snapshot, and Phase 1's non-goals are clearly listed. Good discipline for a foundational PR that could easily have overclaimed.

Approving.

Vishnu2707:
Well Done, approved

PR #273, opened 2026-08-13
feat: add rule AZ-CMP-005 Trusted Launch (Secure Boot + vTPM) check

shariqueahmad108-ship-it:
## What

Implements **Rule 1** of #269 — `AZ-CMP-005`, flagging Generation 2 VMs that don't have **Trusted Launch** (Secure Boot **and** vTPM) fully enabled. (Per the issue's Coordination note, the Managed Identity rule was descoped, so this PR is the sole remaining scope.)

Follows the `scanner/rules/` one-file-per-rule template, with the matching remediation playbook and compliance-framework mappings.

## Detection logic

- **Compliant** → `security_type == "TrustedLaunch"` and both `uefi_settings.secure_boot_enabled` and `v_tpm_enabled` are `True`.
- **Gen1 = NOT_APPLICABLE.** A VM's `list_all()` representation doesn't carry its Hyper-V generation, so the rule resolves the **OS disk's `hyper_v_generation`** (via `azure_client.get_disk()`, the same pattern `AZ-CMP-002` uses) to confirm a VM is Gen2 before flagging. A readable Gen1 disk reports `V1` and is skipped — no false finding against hardware that can't run Trusted Launch.
- **`security_type == "TrustedLaunch"` but Secure Boot / vTPM off** is Gen2-only by definition, so it's flagged **without** needing a disk lookup.
- **Confidential VMs** (Secure Boot + vTPM by construction) are out of scope and skipped.
- **Unknown generation** (OS disk unreadable / missing) is deliberately **left unflagged** — an unreadable disk could be Gen1, and the issue explicitly asks that Gen1 not produce a false finding. This mirrors the repo's existing "never treat an indeterminate result as a violation" stance (`AZ-CMP-002`). Trade-off: a Gen2 VM whose disk generation can't be read is a possible false negative; I erred toward zero Gen1 false positives, but happy to switch to surfacing it as a distinct low-severity/unknown result if you'd prefer the `AZ-CMP-002` treatment.

## Compliance mappings (all four framework JSONs)

| Framework | Control | Rationale |
|---|---|---|
| NIST CSF | `PR.DS-6` | Firmware/software integrity verification (Secure Boot + vTPM measured boot) |
| ISO 27001 | `A.12.5.1` | Installation of software on operational systems — only signed boot code runs |
| SOC 2 | `CC6.8` | Prevents/detects unauthorized software at boot |
[...]

PR #275, opened 2026-08-14
feat: add VM Scale Set inventory collector and rule AZ-CMP-006

### LINKED ISSUE #271: feat: add VM Scale Set inventory, collectors, and first rule

OpenShield currently has zero coverage of Virtual Machine Scale Sets — no inventory collector, no
rules. VMSS is a common production compute pattern and is entirely invisible to the scanner today.

## Describe the solution

- Add a VMSS inventory collector to AzureClient (get_virtual_machine_scale_sets(), following the
  existing collector pattern in docs/adding-a-rule.md).
- Ship one high-value first rule reusing existing compute-domain logic where possible (e.g. a VMSS
  equivalent of AZ-CMP-002's disk-encryption check, or AZ-CMP-001's NSG check applied to the VMSS
  network profile) rather than inventing new detection logic from scratch.
- Add matching playbook, tests, and framework mappings per the standard rule contract.

## Alternatives considered

Wait until AKS/container work (already well underway via issue #255 and 6 existing az_aks rules)
progresses further before touching VMSS — rejected; VMSS is a distinct, currently uncovered gap
that doesn't depend on the container work landing first.

## Additional context

Sequenced before container platform work, per the compute roadmap's own implementation strategy.

dipeshrayg:
## What does this PR do?
Adds the first VM Scale Set coverage to OpenShield: a new `AzureClient.get_virtual_machine_scale_sets()` inventory collector, plus AZ-CMP-005, which flags VMSS network interface configurations with a public IP and no NSG.

## Type of change
- [x] New scan rule
- [x] Remediation playbook
- [x] Compliance mapping

## Rule details
- Rule ID: AZ-CMP-005
- Severity: HIGH
- Category: Compute
- Frameworks mapped: CIS (see note below) / NIST / ISO 27001 / SOC 2

## Design notes
Per the issue's guidance, this reuses AZ-CMP-001's detection logic (public IP + no NSG) rather than inventing new logic, applied to the VMSS network profile template instead of a standalone VM's NIC. A VMSS's `virtual_machine_profile.network_profile.network_interface_configurations[]` embeds `ip_configurations[].public_ip_address_configuration` and `.network_security_group` directly (confirmed against the installed `azure-mgmt-compute` SDK models), so no second collector call (e.g. resolving a NIC by ID) is needed the way AZ-CMP-001 does.

## Testing
- [ ] Tested against a real Azure free trial subscription — not done, no live subscription available
- [x] Returns correct JSON output — `tests/test_rules_compute.py` covers compliant (NSG present), compliant (no public IP), non-compliant, missing-profile, and multi-config cases; `tests/test_azure_client_management.py` covers the new collector's success/failure paths
- [x] All seven CI-equivalent checks reproduced locally and passing: rule syntax, rule structure/duplicate-ID validation, hardcoded-credential scan, playbook existence + bash syntax, compliance JSON validity, API syntax, and compliance↔rule cross-reference
- [x] Full test suite (253 relevant tests) + `ruff check` pass locally

## Related issue
Closes #271

## Note on the CIS mapping
CIS 7.1 ("Ensure that Network Security Groups are attached to network interfaces with public IP addresses") is the same real control, already assigned to AZ-CMP-001 under this repo's one-CIS-ID-per-rule convention. Rather than reuse 7.1 or invent a second number CIS doesn't define, AZ-CMP-005 is mapped to `N/A-CMP-005`, mirroring the pattern already established for AZ-KV-001 (`N/A-KV-001`, referencing AZ-KV-004's 8.5). NIST/ISO27001/SOC2 do reuse AZ-CMP-001's control IDs directly, consistent with `test_cis_benchmark_mapping.py`'s own stated policy that only CIS enforces one-rule-per-control.

## Checklist
- [x] Every commit includes a DCO `Signed-off-by` trailer (`git commit -s`)


TFT444 (changes requested):
@dipeshrayg The VMSS work is useful, but it conflicts with #273 because both PRs use AZ-CMP-005. Please rename this rule to AZ-CMP-006 and update the rule, playbook, tests and all framework keys. Also add the missing SOC 2 entry to the rule's FRAMEWORKS mapping and account for subnet-level NSGs so a protected VMSS is not falsely flagged.

dipeshrayg:
Hey @TFT444 @m-khan-97 — I've pushed the fixes for both of your comments: renumbered to AZ-CMP-006, added the SOC2 mapping, and fixed the subnet-level NSG false positive with new regression tests for it. All CI checks are green. Let me know if there's anything else you'd like changed, happy to take another pass whenever you get a chance to look.

parthrohit22 (dismissed):
Nice piece of work @dipeshrayg , Went through this in detail and ran the full suite against it (564 passed, 3 skipped, clean ruff). Nice work on the NSG check — you're checking both the NIC-level and subnet-level NSG before flagging, which a lot of first-pass rules like this get wrong (would've been a false-positive magnet otherwise). Also double-checked network_security_group, ip_configurations, public_ip_address_configuration etc. against the actual azure-mgmt-compute/network SDK models — all correct, no attribute typos. AZ-CMP-006 is free, doesn't collide with anything else in flight. Approving. keep it up

ritiksah141 (dismissed):

ritiksah141 (changes requested):
There is one correctness issue that can create false findings in `scanner/rules/az_cmp_006.py`.

`AzureClient.get_virtual_networks()` returns `[]` when the API call fails. `_subnet_nsg_map()` therefore produces an empty/incomplete map, and this expression treats every unresolved subnet exactly like a successfully resolved subnet with no NSG:

```python
subnet_nsgs.get(subnet_id, False)
```

That means a VMSS whose subnet really has an NSG will be reported as exposed whenever VNet collection fails, permissions are insufficient, the subnet ID casing differs, or the referenced subnet is otherwise absent from the collected map. This directly reintroduces the subnet-level NSG false positive the latest commit was intended to fix.

Please distinguish “subnet resolved and has no NSG” from “subnet could not be resolved.” An unresolved dependency should not produce a definitive HIGH finding; use the repository's unknown/incomplete-data behavior or safely skip that VMSS/configuration. Normalize Azure resource IDs case-insensitively as well, and add regression tests for an unresolved subnet/VNet collection failure and differently-cased resource IDs.

dipeshrayg:
Good catch, that's a real regression. Fixed: `_subnet_nsg_map` and the lookup now distinguish a resolved subnet confirmed to have no NSG from one that couldn't be resolved at all. An unresolved subnet now produces a LOW-severity indeterminate finding instead of a confirmed HIGH, following the same confirmed/indeterminate pattern az_cmp_002.py already uses. Subnet IDs are also normalized to lowercase before comparison now, since the two APIs involved don't guarantee matching casing. Added regression tests for both cases (VNet collection returning empty, and a differently-cased subnet ID match). All 48 relevant tests pass.

ritiksah141 (dismissed):
Looks good to me, Approving it.

TFT444 (dismissed):
**BLOCKER — False HIGH via non-primary ip_config subnet: `_subnet_nsg_status` iterates ALL ip_configs on a net_config, not just those carrying a public IP.** A non-primary ip_config referencing a subnet with no NSG forces `subnet_statuses` to contain `False`, triggering a confirmed HIGH finding even though the non-primary config is not internet-reachable — producing an exploitable false positive that incorrectly flags compliant VMSSes as exposed.

**BLOCKER — `break` after first offending net_config silently suppresses all subsequent exposed configs.** Only the first non-compliant NIC config name appears in `metadata.network_interface_configuration`; a VMSS with three exposed configs shows one finding, leaving two attack surfaces invisible to remediators — a reporting completeness failure with direct remediation impact.

**MAJOR — VNet collection failure (`get_virtual_networks` returns `[]` on both genuine-empty and API error) silently downgrades every VMSS finding to LOW with no operator signal in the findings.** A persistent permission failure produces only indeterminate LOW findings across the entire subscription indefinitely, with no finding-level escalation path — fix: distinguish error from genuine-empty or surface the degraded-collection state in finding metadata.

dipeshrayg:
Rebased this branch onto current `dev` (it had drifted since 08-19 and picked up a real merge conflict with the JIT-access rule that landed later, AZ-CMP-007). Conflicts were all additive, both rules' entries kept side by side in the four compliance JSON files and in test_rules_compute.py.

One real bug turned up while rebasing: AZ-CMP-007's tests and this PR's AZ-CMP-006 subnet-fix tests each defined their own `_subnet_id()` helper with a different signature. Landing both in the same file let the second definition silently shadow the first and broke one of AZ-CMP-007's tests. Fixed by renaming this PR's helper to `_vnet_subnet_id()`.

Verified after the rebase: 870/870 tests pass, ruff clean, rule structure check clean (107 rules, no duplicate IDs), no more merge conflict. All 20 CI checks are green on the new head.

The force-push reset the review requirement, so this needs a fresh look even though nothing in the original diff changed, just the base and the one collision fix above.

dipeshrayg:
Thanks, both blockers and the major issue were real. Fixed:

- Non-primary ip_config subnet false HIGH: subnet status is now only checked for ip_configurations that actually carry a public IP. A non-primary ip_config with no public IP no longer contributes to the finding.
- Suppressed subsequent exposed configs: removed the `break`. Every non-compliant net_config on a VMSS now produces its own finding.
- Degraded VNet collection with no operator signal: indeterminate findings now carry `vnets_collected` in metadata, so a persistent zero across many findings is visible as a collection problem rather than reading as an ordinary per-subnet indeterminate result. Full error-vs-empty disambiguation isn't possible without an AzureClient change (list methods return `[]` on both today, a repo-wide convention, not something specific to this rule), so this is the scoped fix within the rule itself.

Added regression tests for all three: a non-primary ip_config with an unresolved subnet on an otherwise NIC-protected config produces no finding, a VMSS with multiple exposed configs reports every one of them, and an indeterminate finding's metadata reports `vnets_collected: 0` when VNet collection returns nothing.

873/873 tests pass, ruff clean, rule structure check clean. Pushed.

TFT444 (changes requested):
All three blockers from the original TFT444 CHANGES_REQUESTED (rename to AZ-CMP-006, add SOC2, subnet-level NSG) are resolved. The three issues from TFT444's most recent review (non-primary ip_config false positive, `break` suppressing multi-config findings, VNet failure degradation) are also addressed. One correctness bug remains.

---

## [BLOCKER] `has_subnet_nsg` aggregation creates a false PASS when one public ip_config is protected but another on the same net_config is not

In `scanner/rules/az_cmp_006.py`, lines 246-248:

```python
subnet_statuses = [_subnet_nsg_status(subnet_nsgs, ip_cfg) for ip_cfg in public_ip_configs]
has_subnet_nsg = any(status is True for status in subnet_statuses)
if has_subnet_nsg:
    continue  # protected at the subnet level, compliant
```

`any(status is True ...)` returns `True` if any public ip_config's subnet has an NSG, then skips the entire net_config as compliant. A net_config can carry multiple public ip_configs referencing different subnets; if subnet A has an NSG and subnet B does not, the net_config is silently skipped despite having an exposed second ip_config.

Fix: evaluate subnet protection per ip_config. A net_config is only compliant at the subnet level if **all** public ip_configs have a resolved subnet NSG. `has_subnet_nsg` should use `all` with a narrowing guard, or the loop should descend to the individual ip_config level. A regression test is also needed: two public ip_configs, subnet A with NSG, subnet B without, expecting one finding.

---

## [MINOR] PR title still references AZ-CMP-005

All files are correctly named AZ-CMP-006; the PR title is cosmetic but misleads triage history. Please update via the GitHub UI.

---

## Rule ID collision (resolved)

The rename to AZ-CMP-006 fully resolves the conflict with PR #273 (AZ-CMP-005, Trusted Launch). No collision exists; neither `az_cmp_005.py` nor `az_cmp_006.py` exist on main today.

---

## Items verified as resolved

- Renamed to AZ-CMP-006 throughout (rule file, playbook, framework keys, tests)
- SOC2 `CC6.6` entry added
- Subnet-level NSG lookup implemented via `_subnet_nsg_map` + `_subnet_nsg_status`
- Unresolved subnet correctly returns `None` and produces an indeterminate finding rather than a false PASS
- Case-insensitive subnet ID matching via `.lower()` normalization
- Non-primary ip_config without a public IP excluded from subnet checks
- `break` removed; every exposed net_config emits its own finding
- VNet collection failure surfaces `vnets_collected: 0` in finding metadata

Collector calls `compute_client.virtual_machine_scale_sets.list_all()` correctly. Playbook exists and passes `bash -n`. All 4 compliance framework JSONs contain `AZ-CMP-006` entries. Engine auto-discovers via glob; no explicit registration needed.

dipeshrayg:
Good catch, that was real. `any(status is True for status in subnet_statuses)` let one protected public ip_config wave through a net_config that had another public ip_config still exposed on a different, unprotected subnet.

Fixed: compliance now requires every public ip_config on the net_config to be protected (`all(status is True ...)`), not just one. A confirmed-unprotected ip_config makes the net_config non-compliant even if another ip_config on it is only unresolved, matching the existing confirmed-beats-indeterminate severity model elsewhere in this rule.

Added the regression test you described: two public ip_configs on different subnets, one protected and one not, expects one confirmed HIGH finding.

Also renamed the PR title to AZ-CMP-006.

996/996 tests pass, ruff clean, rule structure check clean (127 rules, no duplicate IDs). Pushed.

dipeshrayg:
Checking in on this one. The subnet-aggregation fix from the last review round has been up for five days now with all 20 CI checks green and no new feedback. Let me know if there's anything else you'd like changed, or if this is ready for another look whenever you get a chance.

Vishnu2707:
Please clear the conflicts in test_rules_compute.py file, post which @TFT444 can re-review this PR!

TFT444 (commented):
@dipeshrayg all logic blockers are resolved and the implementation is correct. There is a merge conflict in `tests/test_rules_compute.py` caused by `dev` drift since Sep 10. No logic changes needed, just a rebase onto current `dev`. Once CI goes green on the rebased head this is ready to merge.

dipeshrayg:
Merged current `dev` in rather than rebasing. `dev` now has its own real AZ-CMP-005 rule from #273, and this PR's own history briefly touched that same path before renaming it to AZ-CMP-006, so a rebase would have replayed that as a false add/add conflict. A merge compares final trees instead, so it only surfaced the one real conflict, in test_rules_compute.py, resolved by keeping both this PR's AZ-CMP-006 tests and #273's AZ-CMP-005 tests side by side, nothing dropped.

1141/1141 tests pass, ruff clean, rule structure check clean (128 rules, no duplicate IDs including AZ-CMP-005 and AZ-CMP-006 now coexisting correctly), all 20 CI checks green, no more merge conflict.

Ready for a final look whenever convenient.

TFT444 (approved):
All blockers addressed: AZ-CMP-006 rename, SOC2 mapping, subnet NSG false positive, unresolved subnet handling, non-primary ip_config exclusion, multi-config reporting, and the 'all' aggregation fix. Merge conflict resolved cleanly. CI green. Approving. Good instinct throughout on this one @dipeshrayg, thanks for the persistence and thorough fixes.

m-khan-97 (approved):
I reviewed the resolved head and the final AZ-CMP-006 behavior. Public IP configurations are now evaluated together correctly: the network configuration is considered protected only when every public IP configuration resolves to a subnet with an NSG, while a confirmed unprotected subnet still produces a HIGH finding and unresolved-only evidence remains indeterminate. The mixed protected/unprotected regression test covers the earlier false pass. I also verified the playbook syntax and ran the compute-rule suite locally: all 75 tests passed. Approved.

Issue #259, opened 2026-08-13
Implement enterprise DevSecOps, supply chain, and AI service rules


ritiksah141:
## Objective

Implement DevSecOps, software supply-chain, and Azure AI service security controls across supported repository providers and Azure resources.

## Scope

Implement these controls:

1. A CI/CD workflow uses long-lived Azure credentials instead of workload identity federation.
2. A CI workflow has unnecessarily broad token permissions.
3. A third-party workflow action is not pinned to an immutable commit.
4. Untrusted pull-request input reaches a privileged workflow context.
5. A protected branch permits unreviewed production changes.
6. A release artifact lacks an approved signature or provenance attestation.
7. Infrastructure deployment can bypass required security scanning.
8. An Azure OpenAI or Foundry resource permits unnecessary public access.
9. An AI service uses static keys where managed identity is available and required.
10. An AI resource lacks required diagnostic logging, encryption, or content-safety controls.

## Collection architecture

- GitHub and Azure DevOps repository, workflow, branch-protection, permission, artifact, and security APIs.
- Repository workflow YAML and deployment configuration analysis.
- Azure OpenAI/Foundry resource, identity, network, diagnostic, encryption, and safety configuration APIs.
- Approved provider and repository scope configuration.

## Required implementation behavior

- Treat unsupported providers, repositories, and optional features as NOT_APPLICABLE.
- Return UNKNOWN for missing repository permissions, inaccessible workflows, unavailable attestations, or Azure API failures.
- Return FAIL only when unsafe configuration is positively evidenced.
- Never expose secrets, tokens, or credential values.
- Include repository/resource context, workflow or branch, observed/expected state, evidence/timestamp, remediation, permissions, severity, confidence, and UNKNOWN reason.

## Delivery order

1. Workflow credentials, token permissions, action pinning, and pull-request context.
2. Branch protection, artifact provenance, and scan-bypass controls.
3. AI network, identity, logging, encryption, and content-safety controls.

## Acceptance criteria

- Tests cover secure/insecure YAML, action pinning, pull-request contexts, branch protection, attestations, scan bypass, AI settings, empty repositories, unsupported providers, malformed YAML, and permission failures.
- Secrets never appear in findings or logs.
- Every rule has documentation, remediation, tests, and verified mappings without TBD.
- Scanner, provider integrations, and CI-equivalent checks pass.
- An origin-branch PR references this issue.

PR #280, opened 2026-08-18
Feat (ci-rules) : Add AZ-CI-001..004 CI/CD workflow security rules (issue #259 PR 1/3)

emon22-ts:
## What does this PR do?
Implements controls 1–4 from issue #259 (DevSecOps and supply-chain security). Adds a GitHub API client, shared workflow YAML helpers, four scanner rules, remediation playbooks, and 46 tests.

## Type of change
- [x] New scan rule
- [x] Remediation playbook
- [x] CI and Testing
- [x] Compliance mapping

## Rules added

| Rule | Severity | Control |
|------|----------|---------|
| AZ-CI-001 | HIGH | Long-lived Azure credentials instead of workload identity federation |
| AZ-CI-002 | HIGH | Unnecessarily broad workflow token permissions |
| AZ-CI-003 | HIGH | Third-party action not pinned to immutable commit SHA |
| AZ-CI-004 | HIGH | Untrusted PR input reaches privileged workflow context |

## New files
- `scanner/github_client.py` — GitHub App installation auth with GITHUB_TOKEN PAT fallback, returns None on missing permissions (UNKNOWN handling)
- `scanner/rules/_workflow_common.py` — shared YAML parsing, action pinning detection, permission analysis, dangerous trigger detection
- `scanner/rules/az_ci_001.py` — long-lived credential detection
- `scanner/rules/az_ci_002.py` — broad token permission detection
- `scanner/rules/az_ci_003.py` — unpinned action detection
- `scanner/rules/az_ci_004.py` — pwn-request pattern detection
- `playbooks/cli/fix_az_ci_001..004.sh` — remediation playbooks
- `tests/test_rules_ci_workflow.py` — 46 tests

## Compliance mappings
- CIS: N/A-CI-001..004 (no verified CIS Azure Benchmark number for CI/CD controls)
- NIST: PR.AC-1, PR.AC-4, PR.IP-1 (evidence-backed per OpenSSF Scorecard, SLSA)
- ISO 27001: A.9.2.4, A.9.2.3, A.12.1.2, A.14.2.5
- SOC2: CC6.1, CC6.3, CC7.1, CC8.1

## Testing
- [x] 46 tests passing — secure/insecure YAML, action pinning, pwn-request patterns, empty repos, unsupported providers, malformed YAML, permission failures
- [x] ruff check and ruff format clean

## Related issue
Refs #259 (PR 1 of 3 - controls 1 - 4)


TFT444 (changes requested):
Thanks solid structure overall, the N/A-CI-* convention is right, tests are thorough, and the NIST mapping fix (PR.DS-5 → PR.AC-4) looks good. All CI is green. However there is one blocking integration bug plus a few correctness gaps that need fixing before merge:

## Blocking

**1. The four rules are loaded by the scan engine but can never run.**
`engine.py` glob-loads every `az_*.py` in `scanner/rules/` and calls `rule.scan(self.client, self.subscription_id)` with an `AzureClient` (engine.py:71, engine.py:114). The AZ-CI rules declare `scan(github_client, owner, repo)`  three required params so **every production scan raises `TypeError` for all four rules**, incrementing `RULE_ERRORS_TOTAL` and logging 4 errors per scan, while the rules silently never execute. Even if the wiring lands in PR 3/3, this PR must either exclude `az_ci_*` from the engine loader or provide an engine-compatible adapter, otherwise every scan between the two merges is polluted.

## Should fix

**2. `github_client.py get_workflows()` does not paginate.**
`/repos/{owner}/{repo}/actions/workflows` returns 30 items per page by default. Repos with >30 workflows are silently truncated → false negatives on exactly the large repos this feature targets. Follow the `Link` header or pass `per_page=100` + loop.

**3. `_workflow_common.py`  `collect_uses()` misses job-level `uses:` and crashes on empty jobs.**
Reusable workflow calls (`jobs.<id>.uses`) are never collected, so AZ-CI-003 can't flag an unpinned reusable workflow. Also `jobs: {build: null}` → `None.get("steps")` → `AttributeError`.

**4. `az_ci_001.py`  OIDC anywhere in the file suppresses the long-lived-credential finding.**
A workflow using `azure/login` with OIDC *and* separately referencing `ARM_ACCESS_KEY` passes clean. The suppression should require the credential to be absent, not merely OIDC to be present. Related: `uses_workload_identity()` matches the literal string `"id-token: write"`, which breaks on extra spacing or quoted YAML — check the parsed permissions dict instead.

## Nits

**5. `github_client.py` get_workflow_content()`:** for files >1 MB the contents API returns an empty `content` field; the code decodes `""` and caches it as valid content, so an oversized workflow scans as compliant instead of UNKNOWN. Treat empty `content` with a present `download_url`/size as a failure.

**6. `tests/test_rules_ci_workflow.py` `test_no_token_returns_none_from_get_token`:** `assert token is None or isinstance(token, str)` is always true; assert `token is None` under the cleared env instead.

emon22-ts:
Hi @TFT444 @Vishnu2707 @H-Sihan , all 6 review items addressed:

1. Blocking - engine.py now excludes az_ci_* from the AzureClient loader with an explicit startswith("az_ci_") guard and debug log
2.  get_workflows() now passes per_page=100 to avoid 30-item truncation
3.  collect_uses() handles job-level uses (reusable workflows) and null job definitions safely
4.  az_ci_001 suppression now requires the specific credential to be absent, not just OIDC to be present; uses_workload_identity() checks parsed permissions dict instead of string-matching
5. Large file handling — empty content with download_url present is treated as UNKNOWN
6.  test_no_token_returns_none now asserts token is None under a fully cleared env

46 tests passing, ruff clean, single signed commit.

Ready for re-review. Thank you!

TFT444 (changes requested):
@emon22-ts good progress: engine exclusion, `collect_uses` hardening, the 1 MB guard, and the token test are all properly fixed. One blocker remains:

**AZ-CI-001 OIDC suppression is still broken, and the new comment claims otherwise.** `az_ci_001.py` says "A workflow using OIDC but also referencing ARM_ACCESS_KEY still fails", but `uses_workload_identity()` checks only `_LONGTERM_CRED_PATTERNS[0]` (`AZURE_CLIENT_SECRET`). OIDC + `ARM_ACCESS_KEY`/`ARM_CLIENT_SECRET`/`AZURE_STORAGE_KEY`/`AZURE_PASSWORD` still passes clean. Make the suppression require no credential match at all (e.g. gate on `not creds` in the scan, or check every pattern), and add a regression test for OIDC + `ARM_ACCESS_KEY`, since the absence of that test is exactly why this slipped.

Smaller items, same push: `get_workflows()` sends `per_page=100` but still doesn't follow the `Link` header (docstring says "Paginates"; >100 workflows still truncate), `uses_workload_identity()` ignores job-level `permissions` so job-scoped `id-token: write` false-positives, and the four framework JSONs lost their trailing newlines.

Fix the AZ-CI-001 suppression and this is mergeable from my side.

scanner/rules/az_ci_004.py
```diff
@@ -0,0 +1,84 @@
...
+)
+PLAYBOOK = "playbooks/cli/fix_az_ci_004.sh"
+logger = logging.getLogger(__name__)
+
+_CHECKOUT_PATTERNS = ("actions/checkout", "git checkout", "git clone")
+
+
+def _checks_out_pr_code(content: str) -> bool:
+    """Return True if the workflow appears to check out PR branch code."""
+    return any(p in content for p in _CHECKOUT_PATTERNS) and (
```
parthrohit22: This only checks whether `actions/checkout` and `github.event.pull_request.head`/`head_ref` co-occur anywhere in the raw file text - it never checks whether the checkout step's `ref:` is actually set to the untrusted value. Repro: a `pull_request_target` workflow that checks out the base ref (safe - no `ref:` override) and only interpolates `github.event.pull_request.head.sha` into a PR comment (exactly GitHub's own documented-safe mitigation) still gets flagged HIGH. `test_safe_prt_without_checkout_no_finding` only covers the trivial no-checkout-at-all case, so this gap isn't caught.

Failure scenario: any label/triage/comment workflow on `pull_request_target` that references `head.sha` for display purposes gets a false HIGH "pwn-request" finding.

scanner/rules/_workflow_common.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Checks the parsed permissions dict for id-token: write rather than
+    string-matching, to handle varied YAML formatting. Also verifies
+    that azure/login is present without a client-secret argument.
+    """
+    has_oidc_permission = False
+    if parsed is not None:
+        perms = parsed.get("permissions") or {}
+        if isinstance(perms, dict):
+            has_oidc_permission = perms.get("id-token", "").lower() == "write"
```
parthrohit22: This only reads the top-level `permissions` block. Job-level `permissions:` (GitHub's own recommended least-privilege pattern) is invisible here. Repro: a `deploy` job with `permissions: id-token: write` set at the job level, using `azure/login` with no client-secret (genuine OIDC), alongside an unrelated `tf-state` job that legitimately uses `ARM_ACCESS_KEY` for a Terraform remote-state backend - `az_ci_001.py` still emits a false HIGH "long-lived credential" finding on the deploy job because `uses_workload_identity` can't see the job-level grant that makes it real OIDC.

scanner/rules/_workflow_common.py
```diff
@@ -0,0 +1,172 @@
...
+    else:
+        has_oidc_permission = "id-token: write" in yaml_content
+
+    has_azure_login = "azure/login" in yaml_content
+    # Check ALL long-lived credential patterns — OIDC suppression requires
+    # every credential to be absent, not just AZURE_CLIENT_SECRET.
+    has_any_long_lived_cred = any(p.search(yaml_content) for p in _LONGTERM_CRED_PATTERNS)
```
parthrohit22: This scans the raw YAML text, including comments. Repro: a workflow with proper top-level `id-token: write` + `azure/login` (no secret) and no long-lived credential actually used, but with a migration comment like `# Old workflow used ARM_CLIENT_SECRET, replaced by federated identity below.` - the credential-name regex matches inside the comment, `has_any_long_lived_cred` becomes true, and az_ci_001 flags it HIGH even though the workflow is genuinely OIDC-only. Any team that leaves a changelog-style comment when migrating to OIDC (a natural thing to do) hits this.

scanner/rules/_workflow_common.py
```diff
@@ -0,0 +1,172 @@
...
+    if perms is None:
+        return None
+    if isinstance(perms, str):
+        return {"all": perms}
+    return perms if isinstance(perms, dict) else None
+
+
+def is_permissions_broad(permissions: Optional[Dict[str, str]]) -> bool:
+    """Return True if permissions grant write-all or multiple broad scopes."""
+    if permissions is None:
```
parthrohit22: An absent top-level `permissions:` block is unconditionally treated as broad here, even when every individual job explicitly restricts its own permissions - which is itself a real, commonly-recommended secure pattern (explicit per-job least privilege instead of a workflow-level default). Repro: a workflow with no top-level permissions block but where every job declares `permissions: contents: read` still gets flagged HIGH by az_ci_002, which directly contradicts that rule's own REMEDIATION text ("Override per-job only where broader access is needed") - the rule's own guidance describes a pattern its own detection logic flags as a violation.

parthrohit22 (changes requested):
Architecturally this is sound: github_client.py matches the codebase's existing try/except -> logger.error -> None-means-UNKNOWN convention exactly, no RULE_ID collisions against the ~90 existing rules, all 4 playbooks exist and pass bash -n, compliance framework JSON additions match the existing schema, and the YAML-1.1 `on:` parsing as boolean True is handled correctly (`workflow.get("on") or workflow.get(True)`) - an easy, common mistake in GitHub Actions tooling that this PR avoids.

But for a compliance scanner, misclassifying a secure config as a violation is a correctness bug, not a nit, and I confirmed 4 real false-positive paths by feeding realistic workflow YAML directly into each rule's scan() rather than just reading the code - all 4 are in code this PR adds. Left inline comments on each with the specific repro.

One more, documented as intentional rather than an oversight (az_ci_003.py's REMEDIATION explicitly calls Docker actions exempt) but worth a decision either way: `is_action_pinned` unconditionally treats every `docker://` reference as pinned, so `uses: docker://alpine:latest` (a mutable tag) is never flagged - confirmed with a direct repro (0 findings). This is a real gap against the rule's own stated threat model, and OpenSSF Scorecard's Pinned-Dependencies check (cited as this PR's own compliance evidence source) does flag it. Worth confirming whether this is deliberate scope-narrowing for 1/3 or should require `@sha256:...` digests for Docker actions too.

Also flagging for visibility, not blocking: `scanner/engine.py` explicitly excludes `az_ci_*` from the automatic rule loader with a comment claiming a separate CI/CD scan entry point invokes them - I grepped the whole repo (api/, scanner/worker.py, everything) and no such entry point exists anywhere yet, on this branch or dev. Not a bug (nothing crashes), but since this is titled "1/3," worth confirming reviewers know this ships tested logic with no way to actually run it in the product until a follow-up wires it up. Separately, TestGitHubClient only covers trivial success/failure - the Link-header pagination in get_workflows, the GitHub App JWT-auth flow in _get_app_token, and the >1MB download_url branch in get_workflow_content have no coverage, worth closing before 2/3 builds on top of this client.

Full suite (after installing this environment's missing azure-*/prometheus-client/python-dotenv packages, an environment gap unrelated to the PR): 758 passed, 3 skipped. tests/test_rules_ci_workflow.py: 47/47 passed - the existing tests don't catch any of the 4 false positives below because none of them construct a workflow that's secure-by-a-more-specific-mechanism-than-the-naive-check (safe pull_request_target via base-ref checkout, job-level OIDC alongside an unrelated job's real secret, a credential name only in a comment, or per-job explicit least-privilege with no top-level default).

TFT444 (dismissed):
good to go

m-khan-97:
@emon22-ts, thanks for the false-positive fixes. Before I re-review them, please rebase onto current `dev` and rerun the complete CI suite; this head is now well behind the integration branch. Please ping me with the new head SHA once it is ready.

emon22-ts:
@m-khan-97 Hi sir , Rebased onto current dev and CI is running. Head SHA: 1a7db10269e6bd44b6812f18eb3d9779e292ef3f

Merge conflicts in all 4 compliance framework JSONs were resolved — upstream controls preserved and AZ-CI-001..004 mappings retained (110 controls each).

@parthrohit22 @TFT444  ready for re-review

m-khan-97 (changes requested):
Emon, the four reported false-positive fixes survived the rebase: PR-head detection is scoped to checkout `ref`, job-level OIDC is recognized, comment-only credential names are ignored, and fully job-scoped permissions no longer fail merely because the top-level block is absent. CI is green.

I found one remaining permission-model false negative in AZ-CI-002. `is_permissions_broad()` treats an absent top-level block as safe whenever every job merely has a non-null `permissions` field; it never evaluates what those job-level values grant. Therefore a workflow where every job declares `permissions: write-all`, or three sensitive `*: write` scopes, returns compliant. Likewise, a harmless top-level `contents: read` plus a job-level `permissions: write-all` is never examined because only the top-level block is passed into the check. Those are exactly the over-broad token permissions this rule claims to detect.

Please evaluate the effective permissions at both levels: flag `write-all`/`write` and the broad-scope threshold in any job override, while preserving the valid per-job least-privilege case. Add regression tests for (1) no top-level block with every job declaring `write-all`, (2) safe top-level permissions overridden by one broad job, and (3) multiple jobs with narrow explicit permissions remaining clean. Then rerun the suite and request rereview.

emon22-ts:
Hi @m-khan-97 Sir , is_permissions_broad() now evaluates effective permissions at both levels:

1. No top-level block with every job declaring write-all → flags correctly
2. Safe top-level permissions overridden by one broad job-level block → flags correctly  
3. Multiple jobs with narrow explicit permissions → remains clean

The fix uses a nested _block_is_broad() helper that checks both write-all strings and 3+ sensitive write scopes, applied to both the top-level block and every job-level override independently.

53 tests passing. Head SHA: 8146a96

Ready for re-review. Thank you.

parthrohit22 (changes requested):
Requesting changes before approval.

The earlier false-positive fixes are present, but three issues still prevent this PR from delivering reliable CI/CD security coverage:

1. **AZ-CI-004 misses a privileged `workflow_run` path.** It detects PR-head references but not checkouts using `github.event.workflow_run.head_sha`. A `workflow_run` job can access secrets and write tokens; checking out code from the triggering untrusted run is therefore a pwn-request risk that this rule currently reports as compliant.

2. **AZ-CI-002 does not calculate effective token permissions.** It treats an absent workflow-level `permissions` block as broad without considering enterprise, organization, or repository defaults. A centrally enforced restricted default can make the workflow read-only, so this produces false HIGH findings. Resolve the configured default permissions, or report `UNKNOWN` when they cannot be read.

3. **The rules are not executable through OpenShield.** `ScanEngine` skips all `az_ci_*` rules and this PR contains no alternative CI/CD scan entry point, target configuration, persistence flow, or `UNKNOWN` evaluation. As submitted, normal OpenShield scans cannot run these controls.

Please add regression tests for the `workflow_run` checkout case and restricted-default permissions case, then wire the rules into a CI/CD scanning path that records PASS/FAIL/UNKNOWN before requesting re-review.

emon22-ts:
Hi @parthrohit22 @m-khan-97 @TFT444 , all 3 blocking items addressed:

1. AZ-CI-004 workflow_run path — added github.event.workflow_run.head_sha and github.event.workflow_run.head_branch to dangerous checkout patterns. Regression test confirms a workflow_run job checking out head_sha flags correctly.

2. AZ-CI-002 UNKNOWN permissions — scan() now calls get_repo_info() to read the repository's default_workflow_permissions. When the default is "read" the finding is suppressed. When it cannot be read (None) an UNKNOWN-marked finding is emitted at MEDIUM severity. Regression tests for both cases added.

3. CI/CD scan entry point — scanner/ci_engine.py adds CIScanEngine which runs all 4 AZ-CI-* rules and returns RuleEvaluation objects with PASS/FAIL/UNKNOWN status, compatible with the existing evaluation contract in scanner/evaluation.py. PASS is emitted for compliant workflows, UNKNOWN for unreadable files or unavailable permissions, NOT_APPLICABLE for repos with no workflows.

63 tests passing, ruff clean.

Ready for re-review. Thank you.

parthrohit22 (changes requested):
Re-reviewed `a8bf1d6` against my three items from the Sep 7 review.

## 1. AZ-CI-004 `workflow_run` checkout path — resolved

`_DANGEROUS_TRIGGERS` includes `workflow_run` and `_PR_HEAD_REF_PATTERNS` now covers `github.event.workflow_run.head_sha` / `head_branch`, so a `workflow_run` workflow that checks out the triggering run's untrusted head is flagged. Regression test present. Good.

## 2. AZ-CI-002 effective permissions — the resolve path is reading the wrong endpoint

`scan()` calls `github_client.get_repo_info()`, which is `GET /repos/{owner}/{repo}`. That response body does **not** carry `default_workflow_permissions` — that field is only returned by `GET /repos/{owner}/{repo}/actions/permissions/workflow` (which needs `administration: read`). So in production `repo_default_permissions` is always `None`:

- the `repo_default_permissions == "read"` suppression branch is dead code, and
- every workflow with no `permissions:` block and non-explicit jobs emits the `MEDIUM` / `effective_permissions: UNKNOWN` finding regardless of the real org/repo default.

`test_..._restricted_default_permissions` only passes because it mocks `client.get_repo_info = lambda: {"default_workflow_permissions": "read"}` — a field the real method never returns, so the test is asserting behaviour that can't happen against GitHub.

Fix: add a dedicated client method that calls `/repos/{owner}/{repo}/actions/permissions/workflow`, return `None` when it 403s (no admin scope), and have the tests mock *that* method. UNKNOWN-on-unavailable is the right fallback; it just needs to actually be reachable via the correct call, and the "default is read → suppress" claim in the PR description needs a code path that can execute.

## 3. Rules still aren't runnable in an OpenShield scan

`scanner/ci_engine.py` defines `CIScanEngine`, but nothing calls it — grep for `CIScanEngine` / `ci_engine` outside the file and its own tests returns nothing. There is:

- no API route and no hook into `POST /api/scans`,
- no CLI entry point,
- no persistence: `run_scan()` returns `List[RuleEvaluation]` and no caller writes those to `rule_evaluations` / through `save_scan()`, so nothing `get_compliance_score()` reads is ever populated,
- no target resolution: `owner` / `repo` are bare constructor args with no config/env wiring.

So "normal OpenShield scans cannot run these controls" is still true — this is a library class with a usage snippet in its docstring, not an entry point. If the wiring is deliberately deferred to a later PR in the 1/5 series, please say so in the PR/issue scope and I'll track item 3 separately instead of blocking on it; as written the PR text ("CI/CD scan entry point — scanner/ci_engine.py adds CIScanEngine") reads as if it closes this, and it doesn't.

## Non-blocking

- `ci_engine.run_scan()` calls each `rule.scan()` (which already iterates every workflow) and then iterates the workflows again per rule — every workflow YAML is parsed twice per rule. The client cache keeps the HTTP cost down, but consider having the engine consume the rule findings it already computed rather than re-walking.
- `_checks_out_pr_code` only inspects `actions/checkout` steps' `with.ref`. A `run:` step doing `git checkout ${{ github.event.workflow_run.head_sha }}` or `gh pr checkout` is missed. The same gap applies to the `pull_request` patterns so it isn't a regression, but worth a follow-up.

63 CI-rule tests pass locally. Happy to re-review once 2 is reading the right endpoint and 3 is either wired up or explicitly scoped out.

TFT444 (changes requested):
@emon22-ts three issues remain:

1. **Double workflow content fetch (raised in prior review)**: `get_workflow_content(path)` is called once per rule per workflow file (4 rules × N files = 4N API calls for the same content). Cache content keyed by path before the rule loop.

2. **`finding_paths` path mismatch**: `finding_paths` is built from `f.get("resource_name")` but compared against `wf.get("path")` (full path like `.github/workflows/ci.yml`). If any rule sets `resource_name` to just the filename, the lookup misses it and emits a false PASS.

3. **Bare `except Exception` in `_get_app_token()`**: swallows `InvalidKeyError` and other diagnostics. Log `exc` so operators can debug auth failures.

emon22-ts:
Hi, all 3 review items addressed and CI is green:

1. Workflow content now cached once per path (4N → N API calls)
2. Findings matched on metadata.workflow_path; fixed related bug where FAIL findings could be misclassified as UNKNOWN
3. _get_app_token now logs exception type for auth debugging (nosemgrep suppression correctly placed)

63 tests passing. Head SHA: 43e5c88

@parthrohit22  @Vishnu2707 @TFT444  ready for re-review

TFT444 (changes requested):
Re-reviewed. The double-fetch is resolved (content cache keyed by path) and the finding_paths key mismatch is fixed. One issue remains: get_repo_info() calls GET /repos/{owner}/{repo}, but default_workflow_permissions is not returned by that endpoint. It is only available from GET /repos/{owner}/{repo}/actions/permissions/workflow (requires administration: read). In production this means repo_default_permissions is always None and AZ-CI-002 silently produces only UNKNOWN findings. Please add a separate get_workflow_permissions() method calling the correct endpoint.

parthrohit22 (changes requested):
Thanks for addressing the earlier workflow parsing, caching, and pwn-request feedback. I’m requesting changes before merge because AZ-CI-002 still cannot determine repository default workflow permissions in production.

The rule reads `default_workflow_permissions` from `GET /repos/{owner}/{repo}`, but that endpoint does not provide this setting. As a result, the restricted-default path cannot execute and workflows without explicit permissions are reported as UNKNOWN regardless of their actual repository setting. Please add a dedicated client method for the GitHub Actions workflow-permissions endpoint, handle unavailable access as UNKNOWN, and update the tests to mock that method.

Also, the Azure scan engine explicitly skips AZ-CI rules, and there is still no route, CLI command, or persisted scan integration that invokes `CIScanEngine`. Please either wire the controls into an executable OpenShield scan flow or explicitly scope that delivery to a later PR so the feature is not presented as scan-ready.

emon22-ts:
Hi @parthrohit22 @TFT444 @Vishnu2707 , both items addressed:

1. Correct endpoint — added GitHubClient.get_workflow_permissions() which reads GET /repos/{owner}/{repo}/actions/permissions/workflow (the only endpoint exposing default_workflow_permissions; requires administration: read). AZ-CI-002 now reads the repo default via this method instead of get_repo_info(), which never returned that field — so restricted defaults are actually detected, and when the endpoint is unavailable (no administration: read) the finding is emitted as UNKNOWN rather than a false HIGH. Tests updated to mock get_workflow_permissions(), plus new client coverage for the value and failure paths.

2. Executable path + explicit scope — added a CLI runner: `python -m scanner.ci_engine --owner <org> --repo <name> [--json]`, which invokes CIScanEngine and exits non-zero on any FAIL. This makes AZ-CI-001..004 runnable today rather than dead code. The persisted-scan storage and API-route integration that invoke CIScanEngine are explicitly scoped to issue #259 PR 3/3 — the engine.py skip comment now points at the CLI runner and names PR 3/3 so nothing is presented as fully product-wired yet.

65 tests passing, ruff clean, rebased on current dev. Head SHA: 951fa2cb7ac54148d59d3934ef871854b245bdb4

Ready for re-review. Thank you.

parthrohit22 (changes requested):
Hey @emon22-ts 
The earlier concerns are addressed: the workflow-permissions endpoint is now correct, and the CI scanner has a runnable CLI entry point.

One required fix remains. has_long_lived_credentials() ignores only full-line comments. Inline YAML comments mentioning retired credentials—for example # old ARM_CLIENT_SECRET migration note—are still scanned as live secrets and can create a false HIGH AZ-CI-001 finding in an otherwise valid OIDC workflow.

Please make the scan comment-aware for inline comments as well, preferably by scanning parsed YAML values, and add a regression test.

TFT444 (changes requested):
@emon22-ts one blocker remains: `has_long_lived_credentials()` strips only full-line comments (`line.lstrip().startswith("#")`). An inline comment on a value line (e.g. `SOME_VAR: "x"  # migrated from ARM_CLIENT_SECRET`) passes through unchanged and triggers a false HIGH AZ-CI-001 finding on a clean OIDC workflow.

Fix: strip the inline comment suffix before regex scanning (split on ` #` and discard the suffix), or switch to scanning parsed YAML values only. A regression test for the inline-comment case is required.

emon22-ts:
Hi @TFT444 @parthrohit22 , all three items addressed:

1. build_vectorstore test — test_build_vectorstore_calls_load_all_documents now patches both VECTORSTORE_DIR and INDEX_PATH to a real tempfile.TemporaryDirectory, runs build_vectorstore() for real, and asserts the bm25_index.json file is actually written. The broad try/except that could have hidden an index-write failure is gone — any write error now fails the test.

2.Compliance route tests — each of the six supported frameworks (cis, nist, iso27001, soc2, ncsc_pqc, enisa_pqc) now asserts get_compliance_score() is called with the exact framework argument and validates the response payload (framework echoed back, required keys present, passed + failed == total_controls). Added an error-result → 500 case alongside the existing invalid → 400 and db-error → 500 cases.

3.cbom.py — the unrelated `from __future__ import annotations` change is removed; the file is now identical to dev (git diff upstream/dev -- api/routes/cbom.py is empty).

Rebased on current dev. Head SHA: 2c245ccff084447d036642e59a4962aa52ad1a8f

Ready for re-review. Thank you.

TFT444 (changes requested):
@emon22-ts good work on the endpoint fix and CLI runner, that was solid. One small fix still needed before this can merge: `has_long_lived_credentials()` only strips full-line comments but not inline ones. A line like `SOME_VAR: "x"  # migrated from ARM_CLIENT_SECRET` still triggers a false HIGH. Fix: split each line on ` #` and scan only the part before the comment. Add a regression test for that case and this is good to go.

parthrohit22 (changes requested):
Re-reviewed `951fa2c`. The two items from my 18 Sep review are done — `get_workflow_permissions()` reads the correct endpoint, and the `python -m scanner.ci_engine` runner makes AZ-CI-001..004 executable today rather than dead code until PR 3/3. Thank you for both.

The inline-comment item from my 19 Sep review is still open, though. `951fa2c` touched `ci_engine.py`, `engine.py`, `github_client.py`, `az_ci_002.py` and the tests — `has_long_lived_credentials()` in `scanner/rules/_workflow_common.py` is unchanged:

```python
non_comment_lines = [line for line in yaml_content.splitlines() if not line.lstrip().startswith("#")]
```

That only drops lines whose *first* non-whitespace character is `#`. A trailing comment on a line that also carries real YAML survives intact and is scanned as if it were live configuration.

I ran it against the branch head to be sure I wasn't describing a theoretical case. A correct OIDC workflow — top-level `id-token: write`, `azure/login@v2`, no secret anywhere — with a single inline comment:

```yaml
client-id: ${{ secrets.AZURE_CLIENT_ID }}  # migrated off ARM_CLIENT_SECRET in 2025
```

```
credentials matched : ['ARM_CLIENT_SECRET']
uses OIDC           : False
AZ-CI-001 would FLAG: True
```

The second line is worse than what I described last time, and it's the reason I'd like this closed before merge. `uses_workload_identity()` doesn't just fail to suppress the finding — it returns `False` outright, because it calls this same comment-blind helper to satisfy its own "no long-lived credential is used outside comments" clause. So one inline comment both defeats the OIDC detection and triggers the finding. The result is a false HIGH on exactly the workflow this rule is meant to reward, and the remediation text tells the maintainer to migrate to OIDC they have already migrated to.

The fix I'd suggest is the one from last time, which also makes this class of bug go away rather than moving the goalposts: scan the parsed YAML instead of the raw text. You already have `parse_workflow()` producing a dict, and PyYAML has discarded every comment by that point regardless of position. Walking the parsed values for the credential patterns needs no comment handling at all — inline, full-line, or inside a block scalar.

If you'd rather keep the text scan, stripping inline comments correctly means respecting quoting (`#` inside a quoted string is not a comment), which is a second bug waiting to happen — hence the preference for the parsed path.

Please add a regression test with the inline-comment case above, asserting both that no finding is produced and that `uses_workload_identity()` returns `True`. Happy to re-review as soon as it's up.

PR #286, opened 2026-08-18
feat: add frontend API timeout and cancellation support

### LINKED ISSUE #282: feat: add request timeout and cancellation support to the frontend API client

The shared `apiFetch()` function in `frontend/src/utils/api.js` does not create a timeout or cancellation controller for normal frontend API requests.

A request that connects but never completes can therefore leave its consumer waiting indefinitely. Although `apiFetch(path, options)` passes supplied Fetch options through, the public API methods generally do not expose a cancellation path, and there is no mechanism that safely combines caller cancellation with a client-level timeout.

The same file's `testConnection()` already uses `AbortController` with a five-second timeout, while the separate AI request client also has timeout logic. Normal API requests do not receive equivalent protection.

## Describe the solution

Add timeout and cancellation support to the shared frontend request layer:

- provide a configurable default request timeout;
- use `AbortController` to cancel requests when that timeout expires;
- always clean up timeout handles after completion or failure;
- safely compose timeout cancellation with a caller-provided `AbortSignal`;
- expose an appropriate cancellation path through API methods where needed;
- distinguish timeout and caller cancellation from ordinary HTTP failures;
- allow operation-specific timeout overrides for legitimately long-running requests;
- add automated tests for successful responses, HTTP failures, timeout, caller cancellation, and cleanup.

Automatic retries should not be enabled indiscriminately. Non-idempotent or state-changing calls, including scan triggers, must not be blindly retried.

## Alternatives considered

- Add timeout logic independently to each page. This would duplicate behavior and make request handling inconsistent.
- Introduce another HTTP-client dependency. The repository already uses the browser Fetch API and `AbortController`, which provide the required primitives.
- Automatically retry every failure. This could repeat state-changing operations and conceal persistent backend failures.

## Additional context

`testConnection()` in `frontend/src/utils/api.js` already demonstrates timeout cancellation with `AbortController`. The validation guide added by PR #134 also records normal frontend request timeout handling as follow-up work. Current `dev` at `4c16d22` has no tests covering timeout or caller cancellation for `apiFetch()`.

SHAURYAKSHARMA24:
## What does this PR do?

Adds a shared 30-second default timeout and caller cancellation support to the frontend API client without introducing automatic retries.

The request layer composes an internal `AbortController` with an optional caller-provided `AbortSignal`, clears timeout handles and caller listeners in `finally`, and accepts a per-call `timeoutMs` override (`null` disables the internal timeout). Public API wrappers forward request options, including custom headers.

Timeouts, caller cancellation, HTTP responses, and network failures use distinct exported error classes and codes:

- `ApiHttpError`
- `ApiNetworkError`
- `ApiTimeoutError`
- `ApiCancellationError`

Scan polling treats individual timeout/network poll failures as transient and continues toward a later terminal backend status. `getScan()` retains the direct-endpoint HTTP compatibility fallback to `/scans`, while avoiding a second request after a timeout/network failure. Explicit caller cancellation still propagates and stops polling when a signal is supplied.

Optional playbook and CVE-summary data preserve graceful fallbacks for HTTP/network/timeout failures while allowing explicit caller cancellation to propagate. Current components do not yet create AbortSignals; this PR provides and verifies the request-layer capability and polling support.

No automatic retry policy was added. State-changing requests such as scan triggers are attempted once, and caller options cannot override the authoritative POST method/body.

## Type of change

- [ ] New scan rule
- [ ] Remediation playbook
- [x] Bug fix
- [ ] API endpoint
- [ ] Documentation
- [ ] Compliance mapping

## Testing

Executed from `frontend/`:

- `node src/utils/api.test.mjs` — 25 passed
- `node src/utils/scanPolling.test.mjs` — 4 passed
- `node src/utils/aiApi.test.mjs` — 9 passed
- `node src/hooks/usePageData.test.mjs` — 8 passed
- `npm run test:severity` — 5 passed
- `npm run test:a11y` — passed
- `npm run test:i18n` — passed
- `npm run lint` — passed with zero warnings
- `npm run build` — passed

Executed from the repository root:

- `node website/test_toEmbedUrl.mjs` — 15 passed

Coverage includes success, HTTP/network/parse failures, timeout and caller-abort classification, already-aborted signals, caller-vs-timeout races, timeout overrides/disable/validation, timer and listener cleanup, auth/custom headers, scan HTTP fallback, transient scan polling recovery, terminal failed scans, optional-data fallbacks, cancellation propagation, and single-attempt POST safety.

## Related issue

Closes #282

## Checklist

- [x] Branch name follows the project convention
- [x] No automatic retry of state-changing requests

### DIFF
--- a/frontend/src/utils/api.js
+++ b/frontend/src/utils/api.js
@@ -14,18 +14,105 @@ const getToken = () => localStorage.getItem('jwt_token');
 const setToken = (tok) => localStorage.setItem('jwt_token', tok);
 
 // ── Core fetch ─────────────────────────────────────────────────────────────
-async function apiFetch(path, options = {}) {
-  const token = getToken();
-  const res = await fetch(`${API_BASE}/api${path}`, {
-    ...options,
-    headers: {
-      'Content-Type': 'application/json',
-      ...(token ? { Authorization: `Bearer ${token}` } : {}),
-      ...(options.headers || {}),
-    },
-  });
-  if (!res.ok) throw new Error(`API ${res.status} ${res.statusText}`);
-  return res.json();
+export const DEFAULT_REQUEST_TIMEOUT_MS = 30000;
+
+export class ApiRequestError extends Error {
+  constructor(message, { code, cause } = {}) {
+    super(message, cause === undefined ? undefined : { cause });
+    this.name = 'ApiRequestError';
+    this.code = code;
+  }
+}
+
+export class ApiTimeoutError extends ApiRequestError {
+  constructor(timeoutMs, cause) {
+    super(`API request timed out after ${timeoutMs}ms`, { code: 'TIMEOUT', cause });
+    this.name = 'ApiTimeoutError';
+    this.timeoutMs = timeoutMs;
+  }
+}
+
+export class ApiCancellationError extends ApiRequestError {
+  constructor(cause) {
+    super('API request was cancelled', { code: 'CANCELLED', cause });
+    this.name = 'ApiCancellationError';
+  }
+}
+
+export class ApiHttpError extends ApiRequestError {
+  constructor(response) {
+    super(`API ${response.status} ${response.statusText}`, { code: 'HTTP_ERROR' });
+    this.name = 'ApiHttpError';
+    this.status = response.status;
+    this.statusText = response.statusText;
+  }
+}
+
+export class ApiNetworkError extends ApiRequestError {
+  constructor(cause) {
+    super('API request failed due to a network error', { code: 'NETWORK_ERROR', cause });
+    this.name = 'ApiNetworkError';
+  }
+}
+
+export async function apiFetch(path, options = {}) {
+  const {
+    timeoutMs = DEFAULT_REQUEST_TIMEOUT_MS,
+    signal: callerSignal,
+    headers,
+    ...fetchOptions
+  } = options;
+  const controller = new AbortController();
+  let abortSource = null;
+  let responseReceived = false;
+  let timeoutId;
+
+  const abortFromCaller = () => {
+    if (abortSource === null) abortSource = 'caller';
+    if (!controller.signal.aborted) controller.abort();
+  };
+
+  if (callerSignal?.aborted) {
+    abortFromCaller();
+  } else {
+    callerSignal?.addEventListener('abort', abortFromCaller, { once: true });
+  }
+
+  if (timeoutMs != null) {
+    if (!Number.isFinite(timeoutMs) || timeoutMs < 0) {
+      callerSignal?.removeEventListener('abort', abortFromCaller);
+      throw new TypeError('timeoutMs must be a non-negative finite number or null');
+    }
+    timeoutId = setTimeout(() => {
+      if (abortSource === null) abortSource = 'timeout';
+      if (!controller.signal.aborted) controller.abort();
+    }, timeoutMs);
+  }
+
+  try {
+    const token = getToken();
+    const res = await fetch(`${API_BASE}/api${path}`, {
+      ...fetchOptions,
+      signal: controller.signal,
+      headers: {
+        'Content-Type': 'application/json',
+        ...(token ? { Authorization: `Bearer ${token}` } : {}),
+        ...(headers || {}),
+      },
+    });
+    responseReceived = true;
+    if (!res.ok) throw new ApiHttpError(res);
+    return await res.json();
+  } catch (err) {
+    if (abortSource === 'timeout') throw new ApiTimeoutError(timeoutMs, err);
+    if (abortSource === 'caller') throw new ApiCancellationError(err);
+    if (err instanceof ApiRequestError) throw err;
+    if (!responseReceived) throw new ApiNetworkError(err);
+    throw err;
+  } finally {
+    if (timeoutId !== undefined) clearTimeout(timeoutId);
+    callerSignal?.removeEventListener('abort', abortFromCaller);
+  }
 }
 
 // ─────────────────────────────────────────────────────────────────────────────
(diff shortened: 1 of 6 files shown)


TFT444 (approved):
Clean, scoped fix with no unrelated changes pulled in. CVE and Dependabot alert are properly referenced and the updated lockfile looks correct. Good to go. Approving.

@ritiksah141 @parthrohit22  please have look into that thanks

TFT444 (dismissed):
Went through the abort logic carefully. The abortSource preference ordering handles the caller-vs-timeout race correctly and the inFlight guard prevents stale state on rapid re-renders. Tests are focused and meaningful. Approving.

frontend/src/utils/api.js
```diff
@@ -274,78 +361,88 @@ export const api = {
...
 
   // ── Single scan  GET /api/scans/:id (falls back to list) ──────────────────
-  getScan: async (scanId) => {
-    try { return await apiFetch(`/scans/${scanId}`); }
-    catch {
-      const data = await apiFetch('/scans');
+  getScan: async (scanId, options = {}) => {
+    try { return await apiFetch(`/scans/${scanId}`, options); }
+    catch (err) {
+      if (!(err instanceof ApiHttpError)) throw err;
```
parthrohit22: This used to be a bare `catch { ... }` that fell back to listing scans on *any* failure. Now it only falls back on `ApiHttpError` and rethrows everything else - including the new `ApiTimeoutError`/`ApiNetworkError`.

`Header.jsx`'s `executeScan` polls this in a loop (`for (let i = 0; i < 75; i++) { ...; const scan = await api.getScan(scanId); ... }`, ~5 minutes at 4s intervals) with a single `try/catch` around the whole loop. A single transient network blip or timeout on any one poll now throws straight out of the loop and ends polling entirely - the user gets a "Scan failed" toast even though the backend scan is still running to completion. Before this PR there was no timeout and any transient error fell back to `/scans` and let the loop continue.

Given `getScan` doesn't override `timeoutMs`, it also now inherits the new default 30s timeout per call, so this isn't just a network-blip edge case - a single slow response during the 5-minute poll is enough to trigger it.

frontend/src/utils/api.js
```diff
@@ -274,78 +361,88 @@ export const api = {
...
   // ── Playbook  GET /api/findings/:id/playbook ───────────────────────────────
-  getPlaybook: async (id) => {
-    try { return normalizePlaybook(await apiFetch(`/findings/${id}/playbook`)); }
-    catch { return { portalSteps: [], cliCommands: [], validationSteps: [], references: [] }; }
+  getPlaybook: async (id, options = {}) => {
+    try { return normalizePlaybook(await apiFetch(`/findings/${id}/playbook`, options)); }
+    catch (err) {
+      if (err instanceof ApiHttpError) {
+        return { portalSteps: [], cliCommands: [], validationSteps: [], references: [] };
+      }
```
parthrohit22: Same narrowing as `getScan` below. `DetailedScan.jsx`'s `selectFinding` calls `await api.getPlaybook(f.id)` with no try/catch, on both mount and click, so a timeout/network error here becomes an unhandled promise rejection with no fallback UI - actually worse than the pre-PR behavior (which returned the empty-arrays fallback) for exactly the failure modes this PR targets.

parthrohit22 (changes requested):
The core apiFetch timeout/abort implementation is genuinely well-built: clearTimeout and the caller-signal listener removal both run unconditionally in a finally covering every exit path (success, HTTP error, network error, timeout, cancellation, JSON-parse error), the timeout-vs-caller-abort race is handled with a first-writer-wins guard so simultaneous timeout+cancel doesn't produce ambiguous error typing, and the test suite genuinely simulates a hanging request via a fake-timer harness rather than just asserting on mocked returns (it also asserts zero pending timers and zero leaked listeners after completion, which is the part most implementations skip). All 11 tests in api.test.mjs pass, and every existing caller of the touched wrapper functions still works with the new optional `options` param.

But alongside the new apiFetch, three existing wrapper functions (getCVESummary, getPlaybook, getScan) had their catch blocks narrowed from "catch anything, fall back gracefully" to "only catch ApiHttpError, rethrow everything else" - and two of those have real call sites that were never updated for the new thrown error types, which is a functional regression in exactly the scenario (long-running operation, imperfect network) this PR is meant to make more robust. Left inline comments on the two consequential ones.

Suggested fix direction: either broaden the catch back to cover transient failures (e.g. `ApiHttpError | ApiNetworkError`, letting only `ApiCancellationError` propagate since that's genuinely caller-initiated) or update the Header.jsx poll loop and DetailedScan.jsx's selectFinding to handle the new error types explicitly. Also worth noting for later: no component actually passes an AbortController/signal into any api.* call yet, so while apiFetch now correctly composes a caller signal with its internal timeout, nothing aborts in-flight requests on unmount/re-trigger today - the race condition the PR title references isn't fixed at any call site yet, just made possible. Not blocking, just flagging so it isn't mistaken for done.

ritiksah141:
@SHAURYAKSHARMA24, its been a week since @parthrohit22 has requested changes and has not been addressed, so please give it a look.

SHAURYAKSHARMA24:
Addressed the transport-error regressions from review. Scan polling no longer treats a transient timeout/network failure as backend scan failure, while explicit caller cancellation still propagates. The playbook and CVE-summary paths preserve graceful handling for HTTP/network/timeout failures without swallowing caller-initiated cancellation. Added wrapper and polling regression coverage, rebased onto current dev, and reran the frontend API, polling, page-data, severity, a11y/i18n, lint, build, and website tests. All CI checks are green on the updated head. Ready for re-review.

parthrohit22 (approved):
I re-reviewed the current implementation against the previous feedback.

The transient timeout and network-error handling is now correctly integrated into scan polling, while graceful fallbacks and caller cancellation behaviour are preserved. The added tests provide meaningful coverage for these paths.

I found no remaining blocking issues.

Approving this PR.

parthrohit22 (approved):

ritiksah141 (approved):

PR #287, opened 2026-08-18
feat: add recoverable dashboard error states

### LINKED ISSUE #281: feat: add recoverable error states to data-driven dashboard pages

Several data-driven dashboard pages start their initial API requests without handling a rejected promise:

- `frontend/src/pages/Discovery.jsx`
- `frontend/src/pages/Prioritization.jsx`
- `frontend/src/pages/DetailedScan.jsx`
- `frontend/src/pages/Compliance.jsx`
- `frontend/src/pages/Drift.jsx`

These pages render a loader while their data state is unset. If an initial request rejects, the state is never updated and the page can remain on that loader indefinitely. `DetailedScan.jsx` also uses an empty findings array as its loading condition, so a successful empty response is not distinguishable from loading.

Users need to be able to distinguish loading, successful empty data, an API/backend failure, and successful populated data. `frontend/src/pages/Monitoring.jsx` already catches its initial request failure and provides a basic user-visible error pattern, although it currently requires a page refresh to retry.

## Describe the solution

Add recoverable initial-load error handling to the affected pages. The implementation should:

- catch failed initial API requests;
- replace the loader with an explicit user-visible error state after failure;
- provide an explicit retry action without requiring a full page refresh;
- keep loading, empty, error, and successful states distinct;
- use a reusable error-state UI where that fits the existing component structure;
- announce errors accessibly;
- add frontend tests for rejected requests and successful retry behavior.

The exact component and state design can follow the repository's existing frontend architecture.

## Alternatives considered

- Keep showing the loader after a rejected request. This hides the failure and offers no recovery path.
- Silently fall back to mock security data. The current frontend explicitly uses backend data without mock fallbacks, and substituting mock posture data could mislead users.
- Show only a temporary toast. An inline recoverable state better represents a page that cannot load its content.

## Additional context

The gap was recorded as follow-up work in `docs/validation/FRONTEND_API_TESTING.md` by merged PR #134, but that validation PR intentionally documented rather than implemented the fix. Current `dev` at `4c16d22` still demonstrates the behavior above.

SHAURYAKSHARMA24:
## What does this PR do?

Adds recoverable loading, error, empty, and populated states to the data-driven dashboard pages.

Discovery, Prioritization, Detailed Scan, Compliance, Drift, and Monitoring now use a small shared `usePageData` hook for their initial request lifecycle. Failed loads render a reusable inline error panel with a `Try again` action; retry returns to the loader and can resolve to populated, empty, or another error state without reloading the browser.

The hook ignores stale results after unmount and prevents repeated clicks from starting duplicate concurrent loads. Its cleanup-aware initial scheduling also prevents React Strict Mode effect replay from launching a duplicate initial request. Compliance derives its active framework after a successful load rather than writing component state from inside an async loader. Detailed Scan no longer treats an empty findings array as “still loading”; a successful zero-findings response displays a dedicated empty state. Existing page-specific empty states remain in use where appropriate.

The error panel uses `role="alert"`, an assertive live region, semantic headings, and a keyboard-accessible button. It displays page-specific safe copy rather than raw exception details.

This PR is independently based on current `dev` and does not depend on #282. Its request callbacks remain compatible with the timeout/cancellation API proposed there, but no stacked branch or automatic retry policy is introduced here.

## Type of change

- [ ] New scan rule
- [ ] Remediation playbook
- [ ] Bug fix
- [ ] API endpoint
- [ ] Documentation
- [ ] Compliance mapping

Not applicable.

## Testing

- [ ] Tested against a real Azure free trial subscription (not applicable to load-state handling)
- [x] All GitHub CI and security checks pass

Executed from `frontend/`:

- `node src/hooks/usePageData.test.mjs` — 8 load-state and retry tests passed
- `node src/utils/aiApi.test.mjs` — 9 existing AI settings tests passed
- `npm run test:a11y` — passed
- `npm run test:i18n` — passed
- `npm run lint` — passed with zero warnings
- `npm run build` — passed

Focused tests cover initial loading, populated success, successful empty data, rejected requests, retry to populated success, retry to empty success, rapid duplicate retry prevention, stale completion suppression after unmount, and React Strict Mode effect replay.

## Related issue

Closes #281

## Checklist

- [ ] My code follows the rule template in CONTRIBUTING.md (not applicable; no scan rule)
- [ ] I added or updated the matching CLI playbook (not applicable)
- [ ] I added or updated all four compliance framework mappings (not applicable)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -616,6 +616,9 @@ jobs:
       - name: Run aiSettings tests
         run: node src/utils/aiApi.test.mjs
 
+      - name: Run dashboard load-state tests
+        run: node src/hooks/usePageData.test.mjs
+
       - name: Run accessibility and internationalization checks
         run: npm run test:a11y && npm run test:i18n
 
--- a/frontend/src/components/shared/ErrorState.jsx
+++ b/frontend/src/components/shared/ErrorState.jsx
@@ -1,0 +1,30 @@
+import { FiAlertCircle, FiRefreshCw } from 'react-icons/fi';
+import Button from './Button';
+
+export default function ErrorState({
+  title = 'Unable to load this page',
+  description = 'Check your connection and try again.',
+  onRetry,
+}) {
+  return (
+    <div
+      role="alert"
+      aria-live="assertive"
+      className="flex flex-col items-center justify-center py-16 px-6 text-center rounded-2xl border border-severity-high/30 bg-bg-primary dark:bg-bg-dark-secondary"
+    >
+      <div className="w-14 h-14 rounded-2xl bg-severity-high/10 flex items-center justify-center mb-4">
+        <FiAlertCircle size={24} className="text-severity-high" aria-hidden="true" />
+      </div>
+      <h2 className="text-base font-semibold text-text-primary dark:text-text-dark-primary mb-1">
+        {title}
+      </h2>
+      <p className="text-sm text-text-secondary dark:text-text-dark-tertiary max-w-md mb-5">
+        {description}
+      </p>
+      <Button onClick={onRetry}>
+        <FiRefreshCw size={14} aria-hidden="true" />
+        Try again
+      </Button>
+    </div>
+  );
+}
--- a/frontend/src/hooks/usePageData.js
+++ b/frontend/src/hooks/usePageData.js
@@ -47,14 +47,23 @@ export function createPageDataLoader(load, dispatch) {
   };
 }
 
+export function schedulePageDataLoad(loader) {
+  let active = true;
+  queueMicrotask(() => {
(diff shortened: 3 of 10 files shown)


TFT444 (approved):
Solid implementation. The stale-result suppression via request-ID counter is the right approach, and the queueMicrotask guard correctly prevents React Strict Mode double-fire. The eight hook tests cover the failure paths I would have asked about. Approving.

ritiksah141 (approved):
Looks good to me.

TFT444 (approved):
Solid implementation. The stale-result suppression via request-ID counter is the right approach, and the queueMicrotask guard correctly prevents React Strict Mode double-fire. The eight hook tests cover the failure paths I would have asked about. Approving.

PR #292, opened 2026-08-20
feat: add enterprise governance and tenant controls

### LINKED ISSUE #257: Implement enterprise governance and tenant control rules
## Objective

Implement governance and tenant-control rules for Azure management hierarchy, policy, RBAC, locks, metadata, provider registration, and configuration drift.

## Scope

Implement these controls:

1. A subscription is outside the approved management-group hierarchy.
2. A required security policy initiative is not assigned at the correct scope.
3. A mandatory preventive policy uses Audit instead of an approved enforcement effect.
4. A policy exemption lacks an owner, justification, or expiration date.
5. A critical production resource lacks an approved deletion lock.
6. A subscription has excessive Owner assignments.
7. Privileged access is assigned at an unnecessarily broad scope.
8. A resource provider is registered without a documented operational requirement.
9. A production resource lacks accountable ownership metadata.
10. Security configuration drift remains unresolved beyond the approved SLA.

## Collection architecture

- Management-group and subscription hierarchy APIs.
- Azure Policy definitions, initiatives, assignments, exemptions, and effective policy state.
- Authorization role assignments and role-definition APIs.
- Resource locks, Resource Graph inventory, provider registrations, tags, and governance metadata.
- Approved exception and SLA sources.

## Required implementation behavior

- Resolve inherited policy and RBAC state before evaluating a child scope.
- Honor valid exemptions, owners, justifications, expiration dates, and documented exceptions.
- Return NOT_APPLICABLE for out-of-scope subscriptions or resources.
- Return UNKNOWN for inaccessible hierarchy, policy, RBAC, lock, tag, or provider evidence.
- Return FAIL only when effective unsafe configuration is confirmed.
- Include effective scope, inherited assignment, exception state, evidence, remediation, permissions, severity, confidence, and UNKNOWN reason.

## Delivery order

1. Hierarchy, policy assignment, enforcement, and exemption rules.
2. Locks, Owner assignments, and broad privileged-scope rules.
3. Provider registration, ownership metadata, and drift rules.

## Acceptance criteria

- Tests cover inheritance, conflicting assignments, valid and expired exemptions, scope boundaries, RBAC inheritance, locks, deleted resources, tags, provider state, API errors, and permission failures.
- Valid parent controls and active exemptions do not create false findings.
- Every rule has documentation, remediation, tests, and verified mappings without TBD.
- Scanner, compliance, and CI-equivalent tests pass.
- An origin-branch PR references this issue.

TFT444:
## What does this PR do?

Adds ten policy-driven enterprise governance rules covering management hierarchy, Azure Policy, exemptions, RBAC, resource locks, provider registration, ownership metadata, and unresolved configuration drift.

## Type of change

- [x] New scan rule
- [x] Remediation playbook
- [ ] Bug fix
- [ ] API endpoint
- [x] Documentation
- [x] Compliance mapping

## Rule details

- Rule IDs: `AZ-GOV-001` through `AZ-GOV-010`
- Severity: HIGH and MEDIUM
- Category: Governance
- Frameworks mapped: CIS, NIST CSF, ISO 27001, and SOC 2

The collector uses read-only ARM and Resource Graph requests with pagination. Organisation-specific expectations are loaded from a strict policy file. Missing or inaccessible evidence is treated as `UNKNOWN`, while `FAIL` requires confirmed unsafe evidence.

## Testing

- [ ] Tested against a real Azure subscription
- [x] Returns correct JSON output in focused tests
- [ ] All seven CI checks pass

Local validation:

- `25 passed` across governance foundation, governance rules, and CIS mapping tests
- `ruff check .` passed
- `ruff format --check .` passed for 301 files
- All 105 rules, 105 reference rows, playbooks, and framework mappings cross-reference successfully
- All governance playbooks passed `bash -n`
- Full suite result: 747 passed, 3 skipped, 1 unrelated local Chroma dependency failure in `tests/test_rag_dependencies.py`

Live Azure validation was not performed. The PR should remain draft until CI and maintainer review confirm the implementation.

## Related issue

Closes #257

## Checklist

### DIFF
--- a/scanner/rules/_governance_common.py
+++ b/scanner/rules/_governance_common.py
@@ -1,0 +1,412 @@
+"""Shared evaluation helpers for AZ-GOV-001 through AZ-GOV-010."""
+
+from __future__ import annotations
+
+import logging
+import os
+import re
+from datetime import datetime, timezone
+from pathlib import PurePosixPath
+from typing import Any, Mapping
+
+from scanner.governance import GovernanceCollector, GovernancePolicy, load_governance_policy
+
+logger = logging.getLogger(__name__)
+POLICY_ENV = "OPENSHIELD_GOVERNANCE_POLICY"
+OWNER_ROLE_ID = "8e3af657-a8ff-443c-a75c-2fe8c4bcb635"
+
+
+def value(item: Any, field: str, default: Any = None) -> Any:
+    if isinstance(item, Mapping):
+        return item.get(field, default)
+    return getattr(item, field, default)
+
+
+def normal(value_: Any) -> str:
+    return str(value_ or "").strip().lower().rstrip("/")
+
+
+def properties(item: Any) -> Mapping[str, Any]:
+    candidate = value(item, "properties", {})
+    return candidate if isinstance(candidate, Mapping) else {}
+
+
+def load_context(azure_client: Any, subscription_id: str, rule_id: str) -> tuple[GovernancePolicy, dict] | None:
+    path = os.environ.get(POLICY_ENV)
+    if not path:
+        logger.warning("%s: %s is unset; result is UNKNOWN", rule_id, POLICY_ENV)
+        return None
+    try:
+        policy = load_governance_policy(path)
+    except (OSError, ValueError) as exc:
+        logger.warning("%s: governance policy is invalid or unavailable: %s", rule_id, exc)
+        return None
+    cache = getattr(azure_client, "_governance_snapshot_cache", None)
+    if cache is None:
+        cache = GovernanceCollector(azure_client.credential, subscription_id).collect()
+        setattr(azure_client, "_governance_snapshot_cache", cache)
+    return policy, cache
+
+
+def _resource_name(resource_id: str) -> str:
+    return PurePosixPath(resource_id).name or "subscription"
+
+
+def finding(
+    spec: Mapping[str, Any], subscription_id: str, resource_id: str, resource_type: str, evidence: dict
+) -> dict:
+    return {
+        "rule_id": spec["id"],
+        "rule_name": spec["name"],
+        "severity": spec["severity"],
+        "category": "Governance",
+        "resource_id": resource_id,
+        "resource_name": _resource_name(resource_id),
+        "resource_type": resource_type,
+        "description": spec["description"],
+        "remediation": spec["remediation"],
+        "playbook": spec["playbook"],
+        "frameworks": spec["frameworks"],
+        "metadata": {
+            "scope": resource_id or f"/subscriptions/{subscription_id}",
+            "effective_scope": evidence.get("assignment_scope", resource_id),
+            "inherited_assignment": evidence.get("inherited_assignment", False),
+            "exception_state": evidence.get("exception_state", "not_exempt"),
+            "unknown_reason": None,
+            "evidence": evidence,
+            "confidence": "HIGH",
+            "permissions_required": spec["permissions"],
+            "evidence_collected_at": datetime.now(timezone.utc).isoformat(),
+        },
+    }
+
+
+def _unknown(rule_id: str, source: str) -> list[dict]:
+    logger.warning("%s: %s evidence is unavailable; result is UNKNOWN", rule_id, source)
+    return []
+
+
+def _production(resource: Mapping[str, Any], policy: GovernancePolicy) -> bool:
+    if normal(value(resource, "type")) not in policy.production_resource_types:
+        return False
+    tags = value(resource, "tags", {})
+    if not isinstance(tags, Mapping):
+        return False
+    lowered = {normal(key): normal(item) for key, item in tags.items()}
+    return lowered.get(policy.production_tag) in policy.production_tag_values
+
+
+def _assignment_scope(assignment: Mapping[str, Any]) -> str:
+    resource_id = normal(value(assignment, "id"))
+    marker = "/providers/microsoft.authorization/"
+    return resource_id.split(marker, 1)[0] if marker in resource_id else resource_id
+
+
+def _effective_policy_effect(definition: Mapping[str, Any], assignment: Mapping[str, Any]) -> str:
+    policy_rule = definition.get("policyRule", {})
+    then = policy_rule.get("then", {}) if isinstance(policy_rule, Mapping) else {}
+    raw_effect = value(then, "effect")
+    if not isinstance(raw_effect, str):
+        return ""
+    parameter_match = re.fullmatch(r"\[parameters\(['\"]([^'\"]+)['\"]\)\]", raw_effect.strip(), re.IGNORECASE)
+    if not parameter_match:
+        return normal(raw_effect)
+    parameter_name = normal(parameter_match.group(1))
+    assigned = assignment.get("parameters", {})
+    if isinstance(assigned, Mapping):
+        for key, parameter in assigned.items():
+            if normal(key) == parameter_name:
+                return normal(value(parameter, "value", parameter))
+    declared = definition.get("parameters", {})
+    if isinstance(declared, Mapping):
+        for key, parameter in declared.items():
+            if normal(key) == parameter_name:
+                return normal(value(parameter, "defaultValue"))
+    return ""
+
+
+def evaluate(spec: Mapping[str, Any], azure_client: Any, subscription_id: str) -> list[dict]:
+    context = load_context(azure_client, subscription_id, spec["id"])
+    if context is None:
+        return []
+    policy, snapshot = context
+    rule_id = spec["id"]
+    subscription_scope = f"/subscriptions/{subscription_id}"
+
+    if rule_id == "AZ-GOV-001":
+        rows = snapshot.get("hierarchy")
+        if rows is None:
+            return _unknown(rule_id, "management-group hierarchy")
+        if not rows:
+            return _unknown(rule_id, "subscription hierarchy row")
+        actual: set[str] = set()
+        for row in rows:
+            chain = value(
+                row, "managementGroupAncestorsChain", value(row, "properties_managementGroupAncestorsChain", [])
+            )
+            if not chain:
+                chain = properties(row).get("managementGroupAncestorsChain", [])
+            for ancestor in chain if isinstance(chain, list) else []:
+                actual.add(normal(value(ancestor, "id", value(ancestor, "name"))))
+        if actual & policy.approved_management_group_ids:
+            return []
+        return [
+            finding(
+                spec,
+                subscription_id,
+                subscription_scope,
+                "Microsoft.Resources/subscriptions",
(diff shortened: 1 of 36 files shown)


ritiksah141 (commented):
All good on my side. 

Approving it.

TFT444:
@m-khan-97 can you please review that pr. Thank you

TFT444:
@ritiksah141 looks like your approval came through as a comment rather than a formal review approval. Could you re-submit it as an approval so the PR can move forward? No code changes since you reviewed it.

scanner/rules/_governance_common.py
```diff
@@ -0,0 +1,412 @@
...
+    if rule_id in {"AZ-GOV-006", "AZ-GOV-007"}:
+        assignments = snapshot.get("role_assignments")
+        if assignments is None:
+            return _unknown(rule_id, "role assignment")
+        if rule_id == "AZ-GOV-006":
+            owners = [
+                item
+                for item in assignments
+                if normal(properties(item).get("roleDefinitionId")).endswith(OWNER_ROLE_ID)
+                and _assignment_scope(item) == normal(subscription_scope)
```
parthrohit22: This only counts an Owner assignment when it's scoped exactly at the subscription (`_assignment_scope(item) == normal(subscription_scope)`). But the collector's `roleAssignments` query uses `$filter=atScope()`, which Azure documents as returning assignments at the given scope **and any scope above it** that applies down - so a management-group-level Owner grant is included in `assignments` here, just with a scope that will never equal `subscription_scope`.

Verified with a direct repro: fed `evaluate()` one subscription-scoped Owner assignment and one Owner assignment whose `id` is rooted at `/providers/Microsoft.Management/managementGroups/corp/...`, against a policy with `maximum_subscription_owners=1`. Result was `[]` - two real effective Owners, zero findings, when it should have flagged with `owner_count: 2`.

Suggest counting any assignment in `assignments` that resolves to Owner and applies to this subscription (i.e. don't filter on exact scope equality here) rather than only ones scoped directly at the subscription - `_assignment_scope(item)` is still useful to report *where* the grant originates in the finding metadata, just not as a filter that excludes inherited grants from the count.

parthrohit22 (changes requested):
Read the collector and the shared evaluator in full before touching individual rules, since all 10 route through them. Genuinely solid work in both: `GovernanceCollector` handles all three pagination shapes correctly (ARM `nextLink`, Resource Graph `$skipToken`, OData `@odata.nextLink`), and it defensively checks that a `nextLink` actually starts with the real ARM endpoint before following it - a real guard against a malformed or redirected pagination link, not something I see everywhere. Every evidence set (hierarchy, policy_assignments, locks, etc.) is independently nullable so one API failure can't corrupt or silently zero out an unrelated evidence set, `load_governance_policy()` fails loud on any missing/malformed field instead of defaulting, and `test_missing_evidence_never_becomes_failure` proving the indeterminate-handling discipline across all 10 rules in one parametrized test is good test design. No RULE_ID collisions, all 10 playbooks exist and pass `bash -n`, all four compliance framework JSONs have complete AZ-GOV-001..010 mappings. Ran the full suite myself in a clean clone: 737 passed, 3 skipped, 0 failed, ruff clean, CI 20/20 green.

One real bug though, confirmed with a repro rather than just reasoning about it - left inline on the specific line.

AZ-GOV-006 (excessive subscription Owners) undercounts real Owner exposure. The collector queries `roleAssignments?...&$filter=atScope()`, and Azure's `atScope()` filter returns assignments made at the queried scope **and any ancestor scope that applies down to it** - a management group grant included, not just direct subscription-level grants. But the AZ-GOV-006 branch only counts an Owner assignment toward the total when it's scoped exactly at the subscription. I fed `evaluate()` two real Owner assignments - one at the subscription, one inherited from a parent management group - against a policy capping Owners at 1, and got zero findings back. Two effective Owners, threshold of one, nothing flagged. `test_gov_006_owner_threshold` only ever constructs subscription-scoped fixtures, so this gap is genuinely untested. For a rule whose whole purpose is bounding effective Owner exposure, silently missing the most common enterprise pattern (Owner granted at a parent management group) is a real false negative, not a style nit.

m-khan-97:
@TFT444, please address Parth’s substantive governance/API review findings first, then rebase onto current `dev` and remove the now-obsolete temporary ChromaDB audit exclusions introduced before #317 merged. Request rereview only after the corrected head passes the full suite.

TFT444:
@parthrohit22 the one bug you confirmed is fixed: AZ-GOV-006 now counts all Owner assignments returned by `atScope()`, including those inherited from parent management groups, not just subscription-scoped ones. Added a regression test with an MG-scoped fixture to pin the behavior. Also removed the now-obsolete chromadb pip-audit exclusions (chromadb was removed in PR #317) and rebased onto current dev. All 22 governance tests pass.

TFT444:
@m-khan-97 @ritiksah141 @parthrohit22  rebase and the requrent fixes are done. Have a Look Please.

thank you

ritiksah141 (dismissed):

parthrohit22 (dismissed):
Confirmed the AZ-GOV-006 fix — `_governance_common.py` now counts every Owner assignment `atScope()` returns instead of filtering to subscription-exact scope, which is the right call since that filter is exactly what was discarding management-group-inherited grants. Ran the new regression test myself: `test_gov_006_owner_threshold` and the parametrized `test_missing_evidence_never_becomes_failure` case both pass.

Approving.

TFT444:
Thank you for approving merging now as every requirement met.

ritiksah141 (approved):

TFT444:
@parthrohit22 @m-khan-97 can you please double check

m-khan-97 (changes requested):
Tanvir, I rechecked ab79b6b and ran the governance foundation/rule tests locally: 26 passed, including the inherited-Owner regression. That earlier fix is in place.

I found a separate credential-boundary issue in `scanner/governance.py`, so I cannot approve yet. Both `_get_all()` and `_post_values()` accept pagination URLs with `next_link.startswith(ARM_ENDPOINT)`. That also accepts a different hostname beginning with the same text. I reproduced both paths using a fake session and fake token: the second request went to `https://management.azure.com.attacker.invalid/next` with the Authorization header attached. Both safety assertions failed. No real credentials or network requests were used; this demonstrates unsafe handling of a supplied pagination URL, not that Azure normally returns such a URL.

Please parse and validate the URL against the exact approved HTTPS origin before attaching credentials, including rejecting userinfo and unexpected ports. Apply the same boundary to both pagination paths, and do not silently return partial evidence as complete when a continuation is rejected. Add regression tests for valid ARM pagination, lookalike hosts, userinfo URLs and rejected continuations. Please also review redirect handling so it cannot silently invalidate that boundary.

Once that is fixed, I can re-run the focused tests and finish the review. The existing green tests do not exercise this case.

parthrohit22 (commented):
Re-reviewed `ab79b6b` at @TFT444's request.

The AZ-GOV-006 inherited-management-group Owner fix from my earlier review is in place (`_governance_common.py` counts every `atScope()` assignment, regression test with an MG-scoped fixture present).

I can confirm @m-khan-97's credential-boundary finding independently — it is real, not theoretical:

`scanner/governance.py` lines 126 and 168:
```python
url = next_link if isinstance(next_link, str) and next_link.startswith(ARM_ENDPOINT) else ""
```
`ARM_ENDPOINT = "https://management.azure.com"` with no trailing `/`, so `startswith` accepts:
- `https://management.azure.com.attacker.example/next` (suffix host)
- `https://management.azure.com@attacker.example/next` (userinfo — the real host is `attacker.example`)
- `https://management.azure.com:8443/...` (unexpected port)

and `session.get/post` then sends `Authorization: Bearer <ARM token>` (from `_headers()`) to that host. Same loose check on the first hop in `_get_all` (`path.startswith("https://")`), though there the input is always internally constructed.

Suggested fix: compare `urlsplit(next_link)` against the approved origin exactly — `scheme == "https"`, `hostname == "management.azure.com"` (or the configured sovereign-cloud host), no `username`/`password`, and `port in (None, 443)` — before assigning `url`. Apply it to both pagination paths, and when a continuation URL is rejected, fail the call (return `None`) rather than returning the partial page as if it were complete. Worth confirming `session` has redirects disabled or same-origin-checked on these calls too, so a 3xx can't move the credentialed request off-origin after the check.

Not adding a separate block since m-khan's covers it; flagging that I verified it and agree it needs to land before merge.

Vishnu2707:
@TFT444 - please clear the blockers as stated by @m-khan-97

TFT444:
@m-khan-97 @parthrohit22 @ritiksah141 @Vishnu2707

Fixed the ARM pagination SSRF in `be9bd4e`: both `_get_all()` and `_post_values()` now validate the continuation URL against the exact ARM origin (scheme, host, no userinfo, no port deviation) and return `None` on rejection. Redirects disabled on both paths. Regression tests added. CI is green. Re-review appreciated.

ritiksah141:
@TFT444, as due to new PR landed on dev there is multiple merge conflicts so please address those and ask for re-review

parthrohit22 (approved):
Re-reviewed `732361e`. The ARM pagination credential-boundary issue I confirmed on 8 Sep is fixed, and the fix is tighter than what I suggested — approving.

`_is_valid_arm_continuation()` compares `urlparse(url).netloc` against `management.azure.com` for exact equality. Because `netloc` carries userinfo *and* port, that single comparison rejects all three vectors I listed, plus the scheme downgrade:

- `https://management.azure.com.attacker.invalid/next` → netloc differs
- `https://user@management.azure.com/next` → netloc is `user@management.azure.com`
- `https://management.azure.com:8443/next` → netloc is `management.azure.com:8443`
- `http://management.azure.com/next` → scheme check

The explicit `username`/`password` checks are belt-and-braces on top of that, which is fine. Both `_get_all()` and `_post_values()` validate before assignment and **return `None`** on rejection rather than handing back the partial page as if it were complete — that was the part I most wanted, since a silently-truncated evidence set is worse than a failed call in a rule that reasons about absence. `allow_redirects=False` on both, so a 3xx can't move the credentialed request off-origin after the check passes.

The parametrized regression tests in `test_governance_foundation.py` cover exactly those URLs for both pagination paths, plus a direct assertion that `allow_redirects` is `False`. Mutating the validator will fail CI.

The AZ-GOV-006 inherited-management-group Owner fix from my first review is still in place.

**One non-blocking observation:** `_post_graph()` doesn't pass `allow_redirects=False`. Its URL is the fixed ARM Resource Graph endpoint and the continuation is a `$skipToken` in the body rather than a URL from the payload, so there's no attacker-controlled destination — but a 307/308 from that endpoint would still re-send the POST, with the `Authorization: Bearer` header, to wherever `Location` points. One keyword argument closes it for consistency with the other two. Not worth holding the PR for.

Deferring to @m-khan-97 on their outstanding review.

Issue #296, opened 2026-08-21
bug: stop the public readiness endpoint from exhausting the database pool


m-khan-97:
## What is wrong

`/ready` is public. Each request creates a `DatabaseManager`, checks out a pooled connection, calls `ping()`, and never closes the manager or binds it to Flask teardown.

With the default pool maximum of 10 connections per process, repeated unauthenticated readiness requests can exhaust the pool and take all database-backed routes down.

Evidence: `api/app.py:22-24,235-244` and `api/models/finding.py:27,114-135`.

## Required work

- Return the connection in a `finally` block or use the request-scoped manager/teardown path.
- Restrict readiness and metrics at the ingress/edge while keeping an appropriate liveness endpoint.
- Expose database pool utilization and exhaustion metrics without publishing sensitive operational data.

## Acceptance criteria

- More than `DB_POOL_MAX_CONN` sequential readiness calls leave pool checkout/return counts balanced.
- Failure paths also return/discard the connection correctly.
- A PostgreSQL-backed integration test reproduces the old leak and proves the fix.
- Deployment health gates use database-aware readiness without making the endpoint an unauthenticated DoS primitive.

parthrohit22:
I'd like to work on this. #306 already fixed the core connection leak (returning the pool connection via `g.db` + the existing `close_db` teardown, with the PostgreSQL-backed integration test proving it) and added a DB-free `/health` liveness probe, so the remaining scope here is the other two acceptance criteria: restricting `/ready`/`/metrics` from being usable as an unauthenticated DoS primitive, and exposing pool utilization/exhaustion telemetry without leaking sensitive operational data. Picking those up now.

PR #306, opened 2026-08-21
fix(api): return readiness connections to the pool

m-khan-97:
## What changed

The public `/ready` probe now registers its `DatabaseManager` on Flask's request context before calling `ping()`. The existing teardown path therefore returns the pooled connection after both a successful check and a database failure.

I also added:

- unit assertions that `close()` runs for both 200 and 503 responses;
- a PostgreSQL-backed regression that makes more requests than a deliberately small pool and checks that every checkout has a matching return;
- the same regression for a failure raised after the connection has already been checked out.

## Why

Previously every readiness request permanently held one pooled connection. With the default ten-connection pool, repeated unauthenticated calls could exhaust a worker and take database-backed routes down.

## Verification

- `PYTHONPATH=/tmp/openshield-testdeps python3 -m pytest tests/test_observability.py tests/test_database_manager_reliability.py -q` — 16 passed, 2 PostgreSQL cases skipped locally because no `DATABASE_URL` is configured
- `ruff check .`
- `ruff format --check .`
- `python3 -m compileall -q api/app.py tests/test_observability.py`

CI provides PostgreSQL and will execute the two integration cases.

Part of #296. Edge restriction and pool-utilization telemetry remain explicit follow-up work in that issue.

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -236,8 +236,11 @@ def create_app() -> Flask:
     def ready():
         """Readiness probe: 200 when the database is reachable, else 503."""
         try:
-            db = DatabaseManager()
-            db.ping()
+            # Register the manager on Flask's request context before pinging.
+            # The existing teardown handler then returns the pooled connection
+            # on both the success and error paths.
+            g.db = DatabaseManager()
+            g.db.ping()
             return jsonify({"status": "ready"}), 200
         except Exception as exc:
             logger.warning("Readiness check failed: %s", exc)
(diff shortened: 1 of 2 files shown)


ritiksah141 (approved):

TFT444 (approved):
Reviewed the fix and it is correct. Binding DatabaseManager to g is the right pattern here so teardown fires on both the 200 and 503 paths, and the pool-exhaustion regression test gives me confidence the root cause is fully covered. Approving, as 2 reviews have successfully done with zero requests for change, merging it now.

Issue #303, opened 2026-08-21
core: harden scan transactions, leases, idempotency, and durable background work


m-khan-97:
## Why this is needed

The worker reuses one database connection indefinitely. A failed PostgreSQL transaction is not rolled back before failure status is written, so one database error can poison the worker loop permanently.

Long scans use a fixed 60-minute stale threshold with no heartbeat, lease owner or fencing token. A second worker can reclaim a still-running scan and both can persist duplicate results. CVE enrichment runs in a Gunicorn daemon thread and can disappear on restart.

Evidence: `api/models/finding.py:160-231,367-450`, `scanner/worker.py:40-101`, and `api/routes/scans.py:160-198`.

## Acceptance criteria

- All transactions rollback on failure and discard/reacquire broken connections.
- Claims use renewable leases with owner, expiry and fencing token.
- Heartbeat and completion updates require the current fencing token.
- Evaluation/finding persistence is idempotent using stable unique keys/upserts.
- Scan admission has per-subscription quotas, one-active-scan deduplication and idempotency keys.
- Enrichment is a durable claimed job with retries, stale recovery and complete pagination.
- PostgreSQL-backed fault-injection tests cover abort, restart, duplicate delivery, lease expiry and two-worker races.
- Metrics include worker heartbeat, oldest queue age, lease age, retry count and last successful complete scan.

SHAURYAKSHARMA24:
i would like to work on this

Issue #263, opened 2026-08-13
feat: persist PASS/FAIL/ERROR/NOT_APPLICABLE per rule per resource, fix compliance score


parthrohit22:
## What problem does this solve?

Two coupled problems, one fix:

1. The scanner only persists failing findings. `scanner/engine.py::run_scan()` collects a flat
   `findings` list — there is no record of which rules ran clean, which errored, or which
   resources were never evaluated.
2. As a direct consequence, `api/models/finding.py::get_compliance_score()` computes control
   status as `status = "FAIL" if rule_id in failed_rule_ids else "PASS"` — any rule that errored,
   whose resource was never observed, or that never ran is counted as PASS by default. This
   inflates every reported compliance score.

Both block every downstream reasoning-layer feature (drift history, campaign detection,
attack-path confidence) and the eight already-open "enterprise rules" issues (#253, #255, #257,
#258, #259, #261, #262, plus closed #256/#260) that already assume a PASS/FAIL/UNKNOWN/
NOT_APPLICABLE contract exists, with no shared engine support for it.

PR #250 ("Azure Resource Graph inventory foundation") already lands Phase 1 of the underlying
evidence model — normalised, tenant-scoped resource snapshots with complete/partial/failed
collection state — but explicitly does not persist rule-outcome data. This issue is Phase 2: the
rule-outcome side, built on top of #250's snapshots.

## Describe the solution

- New table (e.g. `rule_evaluations`) keyed on (scan_id, rule_id, resource_id) with a `status`
  enum: PASS / FAIL / ERROR / NOT_APPLICABLE, plus a free-text reason for ERROR/NOT_APPLICABLE.
- `ScanEngine.run_scan()` writes an evaluation row for every resource a rule was applicable to,
  not just the ones that failed.
- Update `get_compliance_score()` to compute PASS only from persisted successful evaluations in
  this table — remove the absence-implies-PASS default entirely.
- Document the contract in `docs/adding-a-rule.md` so new rules (including the 8 open enterprise
  issues) target it directly instead of inventing ad hoc indeterminate handling (see
  `scanner/rules/az_cmp_002.py`'s own LOW-severity indeterminate workaround for what happens
  without a shared model).

## Alternatives considered

Ship the schema and the compliance-score fix as two separate PRs — rejected; the fix has no
correct implementation without the schema existing first, and reviewing them apart just means
merging a known-broken intermediate state.

## Additional context

[...]

PR #308, opened 2026-08-21
fix(core): enforce severity contract v1

### LINKED ISSUE #295: bug: handle CRITICAL consistently across scoring, APIs, reports, and frontend
## What is wrong

OpenShield currently has two CRITICAL rules, but both scanner and database severity weights omit CRITICAL. A scan containing only critical findings can therefore score 100.

The drift continues through the stack:

- `scanner/engine.py:132-145`: CRITICAL deducts zero.
- `api/models/finding.py:41,471-493`: CRITICAL deducts zero.
- `api/routes/resources.py:63-79`: a CRITICAL-only resource becomes `risk=NONE`.
- `api/routes/prioritization.py:97-168`: CRITICAL receives the default low weight, is excluded from remediation time, and `criticalFindings` is populated from the HIGH count.
- Several frontend filters, badges and charts omit CRITICAL.

## Required work

Create one versioned severity registry/contract consumed by the scanner, persistence layer, API validation, prioritization, resources, monitoring, frontend badges/filters and documentation. Do not repair each copy independently.

## Acceptance criteria

- CRITICAL is the highest ordered severity everywhere.
- An all-critical scan cannot score 100.
- Resources, prioritization, monitoring, charts, filters and summaries show the same critical count.
- Unknown severity values fail validation instead of silently taking a low/default weight.
- Contract tests cover every consumer and prevent future taxonomy drift.
- The score model is coordinated with #263 so incomplete evidence cannot be hidden by severity arithmetic.

Related: #263.

m-khan-97:
## Why

We have two CRITICAL scanner rules today, but the old scoring paths treated CRITICAL as a zero-point deduction. That meant an all-critical scan could still report a score of 100, while resources and prioritization could downgrade the same findings to NONE or LOW.

For an enterprise security product, that is a trust failure. I have treated this as a cross-product data contract rather than patching individual screens.

## What I changed

- Added a versioned severity contract covering IDs, aliases, rank, posture weight, matrix risk, labels, colours, and UI tones.
- Wired the scanner, persistence layer, API validation, AI ordering, resources, prioritization, compliance output, Sentinel ingestion, and frontend to that contract.
- Made unknown severities fail closed. INFORMATIONAL is accepted only as an input alias and is stored and returned as INFO.
- Added a migration that inventories existing values, normalizes supported legacy data, rejects unknown data, enforces the database constraint, repairs completed scan scores, and records score-contract provenance.
- Changed deployment order so the migration-owning API must become live before a worker for the same SHA is created.
- Added contract, scanner, persistence, API, monitoring, UI-wiring, and rollout-order regressions, plus contributor and architecture documentation.

## Verification

- 128 focused backend and contract tests passed.
- Broader local suite: 721 passed, 3 skipped. Three dependency-only modules unavailable on this host were excluded; clean CI will install and exercise the complete dependency set.
- Ruff check and format, Python compilation, and diff checks passed.
- Node 22.22: severity contract tests 5/5, ESLint passed, and the production Vite build completed successfully.
- The frontend mirror is byte-identical to the canonical contract.

## Rollout note

This migration intentionally corrects historical completed scores, so score changes are expected and auditable. Contract provenance stays nullable during rollout: if a legacy worker finishes after the schema change, its old score cannot be mislabeled as v1. The API now has to finish deployment before the new worker is created.

Before production, I want this migration run against a production-sized staging copy and the old workers drained. Downgrading the schema will not pretend the previously incorrect scores were valid.

## Kept out of this PR

- Evidence completeness and UNKNOWN/NOT_APPLICABLE semantics remain in #263.
- Transactionality, retry idempotency, finding ownership, and rollback recovery remain in #303.

Closes #295
Related to #263 and #303

### DIFF
--- a/alembic/versions/d8e4f6a1b2c3_severity_contract_v1.py
+++ b/alembic/versions/d8e4f6a1b2c3_severity_contract_v1.py
@@ -1,0 +1,100 @@
+"""Enforce severity contract v1 and repair historical scan scores.
+
+Revision ID: d8e4f6a1b2c3
+Revises: c7a2e9f1b3d4
+Create Date: 2026-08-21 00:00:00.000000
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+# Revision identifiers, used by Alembic.
+revision: str = "d8e4f6a1b2c3"
+down_revision: Union[str, Sequence[str], None] = "c7a2e9f1b3d4"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+_CONSTRAINT = "ck_findings_severity_v1"
+
+
+def upgrade() -> None:
+    """Normalize known aliases, reject unknown data, constrain and rescore."""
+    op.execute(
+        """
+        DO $$
+        DECLARE invalid_values text;
+        BEGIN
+            SELECT string_agg(value, ', ' ORDER BY value)
+            INTO invalid_values
+            FROM (
+                SELECT DISTINCT UPPER(BTRIM(severity)) AS value
+                FROM findings
+                WHERE UPPER(BTRIM(severity)) NOT IN
+                    ('CRITICAL', 'HIGH', 'MEDIUM', 'LOW', 'INFO', 'INFORMATIONAL')
+            ) invalid;
+
+            IF invalid_values IS NOT NULL THEN
+                RAISE EXCEPTION 'Cannot apply severity contract v1; unsupported values: %', invalid_values;
+            END IF;
+        END $$;
+        """
+    )
+    op.execute(
+        """
+        UPDATE findings
+        SET severity = CASE UPPER(BTRIM(severity))
+            WHEN 'INFORMATIONAL' THEN 'INFO'
+            ELSE UPPER(BTRIM(severity))
+        END
+        WHERE severity <> CASE UPPER(BTRIM(severity))
+            WHEN 'INFORMATIONAL' THEN 'INFO'
+            ELSE UPPER(BTRIM(severity))
+        END
+        """
+    )
+    op.add_column(
+        "scans",
+        sa.Column(
+            "severity_contract_version",
+            sa.Text(),
+            nullable=True,
+        ),
+    )
+    op.execute(
+        """
+        ALTER TABLE findings
+        ADD CONSTRAINT ck_findings_severity_v1
+        CHECK (severity IN ('CRITICAL', 'HIGH', 'MEDIUM', 'LOW', 'INFO'))
+        NOT VALID
+        """
+    )
+    op.execute("ALTER TABLE findings VALIDATE CONSTRAINT ck_findings_severity_v1")
+    op.execute(
+        """
+        UPDATE scans AS scan
+        SET score = GREATEST(
+            0,
+            100 - COALESCE((
+                SELECT SUM(CASE finding.severity
+                    WHEN 'CRITICAL' THEN 20
+                    WHEN 'HIGH' THEN 10
+                    WHEN 'MEDIUM' THEN 5
+                    WHEN 'LOW' THEN 2
+                    WHEN 'INFO' THEN 0
+                END)
+                FROM findings AS finding
+                WHERE finding.scan_id = scan.scan_id
+            ), 0)
+        ),
+        severity_contract_version = '1.0.0'
+        WHERE scan.status = 'completed';
+        """
+    )
+
+
+def downgrade() -> None:
+    """Remove the v1 constraint; corrected historical scores remain corrected."""
+    op.drop_constraint(_CONSTRAINT, "findings", type_="check")
+    op.drop_column("scans", "severity_contract_version")
--- a/sentinel/ingest.py
+++ b/sentinel/ingest.py
@@ -72,7 +72,9 @@ def normalise(raw, scan_id):
     if not isinstance(raw, dict):
         raise ValidationError("each Sentinel finding must be an object")
     scan_id = bounded_string(scan_id, "scan_id", maximum=128, pattern=re.compile(r"^[A-Za-z0-9._:-]+$"))
-    raw_severity = _safe_text(raw.get("severity", "MEDIUM"), "severity", maximum=16)
+    if raw.get("severity") in (None, ""):
+        raise ValidationError("severity is required for every Sentinel finding")
+    raw_severity = _safe_text(raw["severity"], "severity", maximum=16)
     try:
         sev = normalize_severity(raw_severity)
     except SeverityContractError as exc:
--- a/sentinel/ingest.py
+++ b/sentinel/ingest.py
@@ -11,6 +11,7 @@ from pathlib import Path
 import requests
 
 from api.validation import ValidationError, bounded_string, uuid_string
+from openshield.severity import SeverityContractError, normalize_severity, severity_rank
 
 WORKSPACE_ID = os.environ.get("SENTINEL_WORKSPACE_ID", "")
 SHARED_KEY = os.environ.get("SENTINEL_SHARED_KEY", "")
@@ -71,10 +72,11 @@ def normalise(raw, scan_id):
     if not isinstance(raw, dict):
         raise ValidationError("each Sentinel finding must be an object")
     scan_id = bounded_string(scan_id, "scan_id", maximum=128, pattern=re.compile(r"^[A-Za-z0-9._:-]+$"))
-    sev_map = {"CRITICAL": 4, "HIGH": 3, "MEDIUM": 2, "LOW": 1, "INFO": 0}
-    sev = _safe_text(raw.get("severity", "MEDIUM"), "severity", maximum=16).upper()
-    if sev not in sev_map:
-        raise ValidationError("severity must be CRITICAL, HIGH, MEDIUM, LOW, or INFO")
+    raw_severity = _safe_text(raw.get("severity", "MEDIUM"), "severity", maximum=16)
+    try:
+        sev = normalize_severity(raw_severity)
+    except SeverityContractError as exc:
+        raise ValidationError("severity is outside the OpenShield severity contract") from exc
     compliance = raw.get("compliance", {})
     if not isinstance(compliance, dict):
         raise ValidationError("compliance must be an object")
@@ -93,7 +95,7 @@ def normalise(raw, scan_id):
         "RuleId": _safe_text(raw.get("rule_id", ""), "rule_id", maximum=64),
         "RuleName": _safe_text(raw.get("rule_name", ""), "rule_name", maximum=512),
         "Severity": sev.capitalize(),
-        "SeverityScore": sev_map.get(sev, 0),
+        "SeverityScore": severity_rank(sev),
         "Description": _safe_text(raw.get("description", ""), "description"),
         "Remediation": _safe_text(raw.get("remediation", ""), "remediation"),
         "CisControl": _safe_text(compliance.get("cis", ""), "compliance.cis", maximum=128),
--- a/.github/ISSUE_TEMPLATE/new_rule.md
+++ b/.github/ISSUE_TEMPLATE/new_rule.md
@@ -8,7 +8,7 @@ assignees: ''
 
 ## Rule Details
 - Rule ID: AZ-XXX-000
-- Severity: HIGH / MEDIUM / LOW
+- Severity: CRITICAL / HIGH / MEDIUM / LOW / INFO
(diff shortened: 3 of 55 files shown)


ritiksah141 (dismissed):
All good.

TFT444 (changes requested):
Migration d8e4f6a1b2c3 has the same down_revision = 'c7a2e9f1b3d4' as PR #310's migration. These two PRs cannot both merge without Alembic branching. Coordinate with the #310 author so one migration correctly chains off the other before either is merged.

sentinel/ingest.py
```diff
@@ -71,10 +72,11 @@ def normalise(raw, scan_id):
     if not isinstance(raw, dict):
         raise ValidationError("each Sentinel finding must be an object")
     scan_id = bounded_string(scan_id, "scan_id", maximum=128, pattern=re.compile(r"^[A-Za-z0-9._:-]+$"))
-    sev_map = {"CRITICAL": 4, "HIGH": 3, "MEDIUM": 2, "LOW": 1, "INFO": 0}
-    sev = _safe_text(raw.get("severity", "MEDIUM"), "severity", maximum=16).upper()
-    if sev not in sev_map:
-        raise ValidationError("severity must be CRITICAL, HIGH, MEDIUM, LOW, or INFO")
+    raw_severity = _safe_text(raw.get("severity", "MEDIUM"), "severity", maximum=16)
```
parthrohit22: This still defaults a missing `severity` key to `"MEDIUM"` before validation runs. The surrounding logic in this same function was rewritten by this PR to raise on an *unrecognized* severity value, but a *missing* key silently becomes MEDIUM rather than failing loud - which is exactly the "an all-critical scan could still report a passing score" trust problem this PR's own description calls out as the reason for the whole change.

Failure scenario: a hand-edited or corrupted findings payload fed to `sentinel/ingest.py` is missing `severity` on a record that was actually CRITICAL. Instead of the ingest rejecting the record, it ships a fabricated MEDIUM to the customer's Sentinel workspace. `tests/test_sentinel_input_validation.py` only exercises explicit invalid/valid severity strings, never a missing key, so this is currently unguarded.

`api/validation.py`'s `findings_list` handles the same situation by leaving `severity` as `None` rather than coercing it - suggest matching that here (raise/reject on missing, don't default).

parthrohit22 (changes requested):
Migration is genuinely well-built: validate (fail loud via RAISE EXCEPTION on any out-of-set severity) -> normalize -> add column -> add CHECK NOT VALID then VALIDATE CONSTRAINT separately (correct low-lock ordering) -> repair scores only once the constraint guarantees every severity is canonical. Hand-verified the score arithmetic (GREATEST(0, 100 - COALESCE(SUM(deduction),0)) with CRITICAL=20/HIGH=10/MEDIUM=5/LOW=2/INFO=0) against a few example rows and it matches openshield/severity.py exactly, and the contract/JSON/frontend-mirror sync is byte-exact and CI-enforced (sync-severity-contract.mjs --check). scanner/engine.py re-raising SeverityContractError out of the per-rule try/except (rather than silently dropping just that finding) and worker.py marking the scan failed rather than completed is exactly the fail-loud behavior this PR is going for, and it's correct.

One real gap though, left as an inline comment: sentinel/ingest.py still silently defaults a missing severity key to MEDIUM in a function this PR otherwise rewrote to fail loud on bad severities. That's the exact anti-pattern the PR description cites as the whole reason for this change, in a file the PR already touched, so I don't think it should ship as a known gap.

Three more non-blocking notes, no action needed to merge:
- `test_v1_migration_freezes_the_same_ids_and_weights_as_the_contract` only does static text-matching against the migration source (asserts a string like `"WHEN 'CRITICAL' THEN 20"` appears); the actual data-repair SQL only ever runs against an empty DB in CI. Worth a real integration test that seeds a couple of completed scans with mixed severities and asserts the resulting `scans.score`, since that's the riskiest, most novel part of this migration.
- That same test couples a frozen historical migration file to the live `severity.py` contract module - if a future v2 contract changes a weight, this test would fail unless someone edits an already-applied migration, which `docs/severity-contract.md`'s own guidance says not to do. Worth flagging for whoever writes v2.
- This PR is what introduces synchronous throwing on unrecognized severity into more frontend components (SeverityBadge/RiskBadge/severityDefinition), and there's no ErrorBoundary anywhere in the frontend - an unexpected value reaching one of these would blank the whole page rather than degrade. Not a regression from correct behavior, just worth a follow-up ticket.

Separately: `gh pr view` shows this PR conflicting against `dev` right now (unrelated to the code review) - flagging since #310 also touches `scans` via a migration and the two would fork the Alembic revision graph if merged independently; I've noted the coordination needed on that side.

Full suite (Python 3.11, this sandbox's .venv targets 3.14 which can't build psycopg2-binary): 732 passed, 3 skipped, 2 failed - the 2 failures are test_observability.py hitting an unreachable postgres host in this environment, unrelated to this diff. All severity-contract-relevant tests (severity_contract, sentinel_input_validation, prioritization, resources, clean_scan, engine_integration, input_validation, error_exposure, render_deploy_config) pass 99/99, and frontend `npm run test:severity` passes 5/5 including the contract sync check.

m-khan-97:
@parthrohit22, thanks for catching the Sentinel gap. It is fixed in 5937dab: missing, null, and empty severities are now rejected instead of defaulting to MEDIUM, with regression coverage for all three cases. I also rebased onto the latest dev and preserved both #287's dashboard load/error states and the severity-contract behavior during conflict resolution.\n\nVerification: 102 severity-focused tests passed; the broader backend suite passed 750 tests with 5 expected skips. Fresh CI is running. Please re-review once it is green.

parthrohit22 (approved):
The gap I flagged is genuinely closed. Pulled this branch's actual head into a clean clone and verified rather than trusting the summary: `sentinel/ingest.py` now raises `ValidationError("severity is required for every Sentinel finding")` when the key is missing, `None`, or empty string, before it ever reaches `_safe_text`/`normalize_severity` - the exact "an all-critical scan could still report a passing score" anti-pattern this PR set out to eliminate no longer has a gap in the one file it left open last round. `test_normalise_rejects_missing_or_empty_severity` is parametrized on all three real cases (`{}`, `{"severity": None}`, `{"severity": ""}`) and I ran it myself: passes, along with the rest of `test_sentinel_input_validation.py` and all 22 tests in `test_severity_contract.py`. Full suite: 735 passed, 5 skipped, 0 failed. Ruff clean. Migration chain resolves cleanly (`alembic heads` -> single head at `d8e4f6a1b2c3`).

None of my three non-blocking notes from last round needed to hold this up and I'm not asking for them now - just flagging I didn't see them addressed in case they got lost: the migration-integration-test gap (data-repair SQL still only runs against an empty DB in CI), the frozen-migration/live-contract coupling in `test_v1_migration_freezes_the_same_ids_and_weights_as_the_contract`, and the missing frontend ErrorBoundary now that more components throw synchronously on bad severity.

Migration coordination with #310 is already tracked on both sides (you kept `d8e4f6a1b2c3` chained after `c7a2e9f1b3d4` here, and I've got #310 documented to rebase onto whichever of us merges second) - not a new blocker, just confirming it's still the right call now that I can see both PRs' current state.

Approving.

sentinel/ingest.py
```diff
@@ -71,10 +72,13 @@ def normalise(raw, scan_id):
     if not isinstance(raw, dict):
         raise ValidationError("each Sentinel finding must be an object")
     scan_id = bounded_string(scan_id, "scan_id", maximum=128, pattern=re.compile(r"^[A-Za-z0-9._:-]+$"))
-    sev_map = {"CRITICAL": 4, "HIGH": 3, "MEDIUM": 2, "LOW": 1, "INFO": 0}
-    sev = _safe_text(raw.get("severity", "MEDIUM"), "severity", maximum=16).upper()
-    if sev not in sev_map:
-        raise ValidationError("severity must be CRITICAL, HIGH, MEDIUM, LOW, or INFO")
+    if raw.get("severity") in (None, ""):
+        raise ValidationError("severity is required for every Sentinel finding")
```
parthrohit22: Confirmed this is the actual fix for my original finding, not just claimed - read it directly: missing/null/empty severity is rejected right here, before `normalize_severity()` is even called, so it can't fall through to a silent default. Matches `test_normalise_rejects_missing_or_empty_severity`'s three parametrized cases, which I ran myself.

alembic/versions/d8e4f6a1b2c3_severity_contract_v1.py
```diff
@@ -0,0 +1,100 @@
...
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+# Revision identifiers, used by Alembic.
+revision: str = "d8e4f6a1b2c3"
+down_revision: Union[str, Sequence[str], None] = "c7a2e9f1b3d4"
```
parthrohit22: This is where the fork with PR #310's migration lives - both chain off `c7a2e9f1b3d4`. Not something fixable from this side alone (confirmed by trying the equivalent on my own PR's migration - pointing down_revision at an unmerged sibling's revision file breaks that PR's own CI outright). This needs a merge-order call: whichever of #308/#310 merges second rebases onto the new head and repoints this. Not blocking my approval - flagging for whoever merges.

parthrohit22 (approved):
Re-reviewed from scratch on the current head - the code hasn't changed since my last approval (`5937dab` is both the commit I approved and today's head), but you re-requested review and I'm not going to just rubber-stamp my own old finding without checking, so I re-verified everything myself rather than trusting memory.

**What I re-ran and re-checked just now:**
- All 70 of this PR's own new/changed tests (`test_severity_contract.py`, `test_sentinel_input_validation.py`, `test_engine_integration.py`, `test_clean_scan.py`, `test_prioritization.py`, `test_resources.py`, `test_input_validation.py`) - all pass.
- Full backend suite: 726 passed. (One unrelated failure locally - `test_chroma_default_embedding_uses_onnx_runtime` - is this branch's own stale copy of `tests/test_rag_dependencies.py` from before #317 removed chromadb; not a defect in this PR's own diff, just a symptom of the branch being unrebased since before #316/#309/#317 landed on `dev`. See the rebase note below.)
- `frontend/src/utils/severity.test.mjs`: 5/5 pass.
- Read the actual `sentinel/ingest.py` fix directly (not just the test result) to confirm my original finding is really closed: missing/empty severity is rejected before normalization is even attempted (line 75-76), and anything outside the contract is rejected via `normalize_severity()` + `SeverityContractError` (line 78-81). Matches the test coverage exactly - confirmed, not just claimed.
- Read the full migration (`d8e4f6a1b2c3_severity_contract_v1.py`) line by line. This is genuinely careful work: a pre-flight `DO $$ ... RAISE EXCEPTION` block that refuses to proceed if any existing severity value can't be mapped (fails loud instead of silently corrupting data), the `ADD CONSTRAINT ... NOT VALID` + `VALIDATE CONSTRAINT` pattern (avoids a long-lived table lock on a large `findings` table - the right way to add a CHECK constraint in production Postgres, not the naive one-shot `ALTER TABLE ADD CONSTRAINT`), and a rescore of historical completed scans using the exact same weights as `contracts/severity.v1.json` (verified: 20/10/5/2/0 match exactly). The downgrade explicitly documents that corrected historical scores are not reverted, rather than leaving that ambiguous.

**The one thing still open, and it's not something this PR can fix alone:**

TFT444's `CHANGES_REQUESTED` is still active and still correct: `d8e4f6a1b2c3` and PR #310's migration both chain off `c7a2e9f1b3d4`, and merging both as-is forks the Alembic revision graph. I can't approve that away, and neither PR can unilaterally resolve it - I tried pointing my own PR's migration at an unmerged sibling's revision file and it breaks that PR's own CI (`KeyError` on a revision that doesn't exist outside the sibling's branch), so the fix genuinely requires a merge-order decision: whichever of #308/#310 lands second rebases onto the new `dev` head and repoints `down_revision`. That's a maintainer call on ordering, not a code change here. I've already added a CI check on my side (#310) that asserts `alembic heads` returns exactly one head, so if this does get missed at merge time, it fails loudly instead of silently forking.

Also worth flagging since I checked it directly: this branch's merge-base with `dev` predates #316, #309, and #317 - it's meaningfully behind. Doesn't block my approval (the code itself is correct and tested), but it should be rebased before/at merge so CI reflects current `dev` rather than a stale pip-audit/test snapshot.

Approving - the actual code is solid and my original findings are genuinely fixed. The Alembic coordination is real but is a merge-order decision for a maintainer, not a defect in this PR.

ritiksah141 (approved):

Vishnu2707 (approved):
Approved. LGTM!

Vishnu2707:
@TFT444 - Please approve and merge if everything looks good from your side.

TFT444 (approved):

PR #309, opened 2026-08-21
fix(platform): enforce container runtime gate

### LINKED ISSUE #299: fix(platform): make the distributed image bootable and deployment verification non-skippable
## What is wrong

Several independent mismatches allow deployment to look healthy without proving the product works:

- `Dockerfile:26` launches `api.app:app`, but the WSGI object is named `application`.
- CI builds and scans the image but never starts it.
- Compose has no scanner worker, uses Node 20 while the frontend requires >=22.22, and sets the wrong frontend API variable.
- Deployment smoke preflight expects `AZURE_CLIENT_SECRET`, while the smoke step does not provide it or perform the documented OIDC login.
- The smoke suite silently skips the real scan, does not wait for worker completion, accepts some 500/timeouts, and can pass with skips.
- Health gates use `/health`, not database-aware `/ready`.
- `.dockerignore` excludes `.git` but not this repository's `.gitdata`.

## Required work

Repair the runtime contract and introduce one mandatory staging-to-production verification path covering API, database and worker.

## Acceptance criteria

- The exact published image starts as a non-root user and reaches `/ready`.
- The build context excludes `.gitdata` and other development/history material.
- Compose has an explicit local-only profile with compatible Node, correct API variables and a worker.
- Required smoke tests fail on any skip.
- OIDC or the intentionally selected credential path is implemented consistently in workflow and docs.
- Smoke enqueues one real scan, waits for terminal completion, and verifies persisted evaluation coverage.
- Worker heartbeat and queue age are part of readiness/promotion.
- The final CI gate includes the image runtime and container vulnerability result.

m-khan-97:
## Why

The repository already published a Docker image, but its default command pointed at a WSGI object that does not exist. CI built and scanned that image without ever starting it, so the green container check did not prove the API could boot.

Local Compose had the same kind of contract drift: no worker, PostgreSQL 15, unsupported Node 20, and an API variable the frontend never reads.

## What I changed

- Made the image use the same migration-aware startup path as deployed APIs. Alembic must succeed before Gunicorn starts the real `api.app:application` object.
- Kept the runtime non-root and excluded this checkout's alternate `.gitdata` history from the build context.
- Replaced the optional Dockerfile scaffold with a required runtime gate: build one commit tag, start that tag against PostgreSQL 16, require exact `/ready` JSON, verify the actual runtime UID is non-root, and scan the same tag with Trivy.
- Kept Trivy independent enough to run when a successfully built image later fails its boot check, and made the final CI summary fail on the container result.
- Rebuilt Compose as an explicit `local` profile with loopback-only ports, migration-aware API startup, a worker that waits for API readiness, Node 22.22, `npm ci`, writable dependency/cache volumes, and the `VITE_API_URL` the browser actually uses.
- Aligned Terraform's Vercel variable name and corrected local runtime documentation.
- Added static contract tests so the image, Compose, Terraform, workflow, and final gate cannot silently drift apart again.

## Verification

- Runtime/configuration tests plus deployment config tests: 25 passed.
- Broader local backend suite: 747 passed, 5 expected skips (the Chroma runtime dependency test was excluded locally because Python 3.12 cannot build its pinned native extension on this host).
- Fresh GitHub CI passed the authoritative image build, Alembic startup, exact `/ready` response, non-root runtime UID, Trivy scan, backend coverage, Terraform validation, and final CI summary.
- Repository-wide Ruff and format checks passed.
- Compose and GitHub workflow YAML parsed successfully; `startup.sh` passed `bash -n`.

This host does not have Docker installed, so I am not presenting a local image run as evidence. The new draft PR job is the authoritative boot, migration, readiness, runtime-UID, and Trivy check.

## Merge order

#306 is merged and this branch has been rebased onto the updated `dev`. The recurring Compose `/ready` probe now uses the corrected database-connection lifecycle.

## Remaining #299 work

This is the first runtime slice and does not close #299. Follow-up work still has to prove the exact published GHCR digest, make the real Azure staging scan non-skippable, align OIDC/credential handling, wait for worker completion and persisted evaluation coverage, and add worker heartbeat/queue-age promotion signals.

Part of #299
Dependency #306 satisfied

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,5 @@
 .git
+.gitdata
 .github
 .pytest_cache
 .ruff_cache
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -508,37 +508,98 @@ jobs:
           path: sbom.cyclonedx.json
           retention-days: 90
 
-  # ── Container image scan (Trivy) — scaffolded ahead of INFRA 1 (#154) ─────
+  # ── Container runtime contract + image scan (Trivy) ───────────────────────
   container-scan:
     name: Container Scan (Trivy)
     runs-on: ubuntu-latest
+    env:
+      IMAGE_TAG: openshield-ci-scan:${{ github.sha }}
+      CONTAINER_NAME: openshield-ci-runtime
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: ci
+          POSTGRES_PASSWORD: ci
+          POSTGRES_DB: ci_db
+        ports:
+          - 5432/tcp
+        options: >-
+          --health-cmd "pg_isready -U ci -d ci_db"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
     steps:
       - name: Checkout repository
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
 
-      - name: Check for Dockerfile
-        id: dockerfile_check
-        run: |
-          if [ -f "Dockerfile" ]; then
-            echo "found=true" >> "$GITHUB_OUTPUT"
-            echo "path=Dockerfile" >> "$GITHUB_OUTPUT"
-          elif [ -f "api/Dockerfile" ]; then
-            echo "found=true" >> "$GITHUB_OUTPUT"
-            echo "path=api/Dockerfile" >> "$GITHUB_OUTPUT"
-          else
-            echo "found=false" >> "$GITHUB_OUTPUT"
-            echo "No Dockerfile yet (INFRA 1 / #154 not merged) — nothing to scan, skipping."
-          fi
-
       - name: Build image
-        if: steps.dockerfile_check.outputs.found == 'true'
-        run: docker build -t openshield-ci-scan:${{ github.sha }} -f "${{ steps.dockerfile_check.outputs.path }}" .
+        id: build
+        run: docker build --tag "$IMAGE_TAG" --file Dockerfile .
+
+      - name: Start image
+        env:
+          POSTGRES_PORT: ${{ job.services.postgres.ports[5432] }}
+        run: |
+          set -euo pipefail
+          JWT_SECRET="$(python3 -c 'import secrets; print(secrets.token_urlsafe(32))')"
+          docker run --detach \
+            --name "$CONTAINER_NAME" \
+            --network host \
+            --env "DATABASE_URL=postgresql://ci:ci@127.0.0.1:${POSTGRES_PORT}/ci_db" \
+            --env OPENSHIELD_ENV=production \
+            --env OPENSHIELD_PUBLIC_DEMO=false \
+            --env ALLOWED_ORIGINS=http://127.0.0.1:8000 \
+            --env PORT=8000 \
+            --env "JWT_SECRET=$JWT_SECRET" \
+            "$IMAGE_TAG"
+
+      - name: Verify readiness and non-root runtime
+        run: |
+          python3 - <<'PYEOF'
+          import json
+          import time
+          import urllib.error
+          import urllib.request
+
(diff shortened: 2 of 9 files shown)


TFT444 (commented):
This PR is still a draft and is explicitly blocked by #306. Please merge #306 first, rebase this branch off the updated dev, then mark it ready for review.

parthrohit22 (approved):
Traced the actual bug: `api/app.py` only defines `application` (built inside a guarded `if` block), never a module-level `app` — the old `CMD ["gunicorn", ..., "api.app:app"]` would `AttributeError` at container boot. `CMD ["./startup.sh"]` execs `alembic upgrade head` then `gunicorn ... api.app:application`, matching the real symbol. Confirmed the container-scan job is now actually wired into the required gate end-to-end: services list → summary table → final failure condition — closing the actual gap in the title (previously this job could build and boot-fail without ever failing CI, since it wasn't part of the gate).

Verified a few things that could have quietly gone wrong and didn't:
- `/ready`'s exact-match assertion in CI (`payload == {"status": "ready"}`) matches what the route actually returns.
- `_resolve_jwt_secret()`'s production-strictness path is genuinely exercised — CI sets `OPENSHIELD_ENV=production` with a real 43-char generated secret, not the insecure default.
- The Terraform variable rename (`VITE_API_BASE_URL` → `VITE_API_URL`) is a real fix, not cosmetic — grepped the frontend and confirmed `import.meta.env.VITE_API_URL` is what `api.js`/`aiApi.js` actually read; the old var was silently unused.
- `worker`'s compose dependency on `api: condition: service_healthy` (rather than `db`) is the right choice — it transitively guarantees migrations have run before the worker touches the schema, since the api healthcheck can't pass until `startup.sh`'s migration step succeeds.
- Grepped for other `docker compose` invocations before trusting the new `--profile local` requirement — nothing else silently breaks from it.

New `tests/test_container_runtime_config.py` (7 tests) is static/contract-style rather than an actual container boot, which the PR description is upfront about — and it directly guards the regression (`assert "api.app:app " not in startup`) rather than just asserting the happy path. Full suite passes; the two `test_observability.py` failures I hit are from this sandbox having no reachable `postgres` host and are unrelated to this diff.

Two non-blocking notes, not asking for changes:
- The `/ready` CI check requires the payload to equal exactly `{"status": "ready"}` — any future legitimate addition to that response (timestamp, version) will break this gate even on a healthy API. Worth knowing, not a problem today.
- `.dockerignore`'s `.gitdata` entry doesn't match anything in this checkout — harmless dead config, looks like leftover local tooling.

parthrohit22:
@TFT444 @ritiksah141  ,
I've reviewed latest changes , it looks mergeable , react on this comment with your views. if agreed , We can merge this.

ritiksah141 (approved):
Look good to me. Approving it

m-khan-97:
Rebased onto the latest dev at 50dfcd7 after #287 merged. The rebase was clean, and the focused runtime/deployment contract suite still passes 25/25 with startup.sh syntax validation clean. Fresh CI is running, including the authoritative container startup/non-root/Trivy gate. Previous approvals may be dismissed by the rewritten commit, so I have requested re-review after the checks complete.

parthrohit22 (approved):
@m-khan-97 acknowledged - re-reviewing on the current commit rather than relying on the pre-rebase approval.

Verified this is genuinely a clean rebase, not just taking that on faith: pulled the current diff fresh and it's identical to what I approved before - same 9 files, same +314/-36, and I specifically re-checked the pieces that mattered (`api.app:application` in the startup path with `api.app:app ` still absent, the `container-scan.result != 'success'` wiring into the final gate condition, `VITE_API_URL` in both `docker-compose.yml` and `infra/terraform/vercel.tf`, `worker`/`frontend` both on `api: condition: service_healthy`) - all byte-for-byte the same as before. CI is green across all 20 checks on this exact head, including the Terraform validate/plan job.

Nothing here changes my original assessment. Re-approving on `50dfcd7`.

Issue #294, opened 2026-08-21
security: remove the public demo bearer and establish real authorization boundaries


m-khan-97:
## Why this is a release blocker

The production dashboard currently embeds a JWT with no expiry. The documented generator signs that token with `JWT_SECRET`, and the frontend copies it into localStorage. The API verifies the HS256 signature but does not enforce role, tenant, issuer, audience, or an authorized subscription.

That means the advertised `viewer` token can call write endpoints once the suspended API is restored. The current schema and repository methods are also globally scoped, so the service cannot safely serve more than one organization.

Audit evidence: `frontend/src/App.jsx:16-24`, `scripts/generate_demo_jwt.py:3-38`, `api/app.py:160-191`, `api/routes/scans.py:62-93`, and `alembic/versions/b3f1a2c4d5e6_baseline_schema.py:23-80`.

### Immediate containment

- Keep the API suspended until the exposed token is removed.
- Remove `VITE_JWT_TOKEN` and the automatic `dev-local-token` bootstrap.
- Rotate `JWT_SECRET` and audit any available access logs.
- Do not put bearer credentials in Vite build-time variables or localStorage.

### Durable boundary

- Adopt Entra/OIDC Authorization Code + PKCE (or another approved enterprise IdP flow).
- Validate issuer, audience, subject, expiry and tenant using JWKS.
- Define viewer, operator and administrator policies and enforce them server-side.
- Persist organization/tenant, requester and authorized subscription ownership.
- Require tenant/subscription context in every repository query; evaluate PostgreSQL RLS.
- If the first enterprise release is intentionally single-tenant, enforce a configured subscription allowlist and document that boundary.

## Acceptance criteria

- No credential or reusable session secret is present in public JS.
- Expired, wrong-issuer, wrong-audience and role-inappropriate tokens are rejected.
- A user cannot read or trigger work for an unauthorized subscription.
- Cross-tenant integration tests cover scans, findings, compliance, resources, drift, AI and enrichment routes.
- Secret rotation and incident-containment steps are documented and exercised.
- The API is not restored until the containment checklist is complete.

Related: #243.

parthrohit22:
I'd like to work on this, scoped explicitly.

[...]

PR #319, opened 2026-08-28
fix(api): rate-limit /ready and /metrics, expose db pool telemetry

### LINKED ISSUE #296: bug: stop the public readiness endpoint from exhausting the database pool

parthrohit22:
## What does this PR do?
Closes the remaining scope of #296 (readiness-endpoint pool exhaustion): rate-limits the unauthenticated `/ready` and `/metrics` probe/scrape endpoints per source IP, and exposes database connection pool utilization/exhaustion telemetry on `/metrics`.

## Type of change
- [x] Bug fix
- [x] API endpoint
- [x] Documentation

## Scope note

#306 already fixed the core connection-checkout leak (`g.db` + the existing `close_db` teardown handler) and added `/health` as a DB-free liveness probe, with a real PostgreSQL-backed integration test (`test_ready_returns_every_real_connection_to_the_pool`) proving checkout/return counts stay balanced across repeated readiness probes, including on the failure path. That work is not touched or duplicated here. This PR picks up the two acceptance criteria from #296 that were still open after #306:

1. **Restrict `/ready`/`/metrics` at the edge, while keeping an appropriate liveness endpoint.** `/health` already stays public and DB-free (unchanged). Render's Blueprint format has no path-based access control, so this repository can't express "restrict this path" in `render.yaml` itself — documented that in `docs/deployment/render.md` under a new "Restricting probe/scrape endpoints at the edge" section, with the operational recommendation for whoever runs a reverse proxy/CDN/WAF in front of a real deployment. As the in-app backstop, added `api.observability.probe_rate_limit`: an in-memory, per-process, per-source-IP rate limiter, wired onto `/ready` (5 requests/10s — half the default `DB_POOL_MAX_CONN`) and `/metrics` (20/10s, generous for normal Prometheus scrape intervals). Deliberately *not* the existing Postgres-backed `api.rate_limit.rate_limit`: that does its own DB round trip per check, which would add database load to the exact endpoint meant to protect the database from overload, and risks checking out a second, separately-tracked pooled connection under its own `g.db` alongside the one `/ready`'s own handler manages — only one of which the teardown handler would return. The rejection happens before the view body runs, so a request over budget never reaches the database work it would otherwise trigger.
2. **Expose pool utilization/exhaustion telemetry without publishing sensitive operational data.** Added `DatabaseManager`'s sibling `get_pool_stats()` in `api/models/finding.py`, returning only in-use/idle/max-connection counts and a utilization percentage — never the DSN, host, or credentials. Wired into three new Prometheus gauges (`openshield_db_pool_connections_in_use/idle/max`) refreshed on every `/metrics` scrape via a provider callback registered from `api/app.py`, keeping `api/observability.py` free of project imports (it's reused by the worker, per its own module docstring).

## Testing
- [x] All CI checks pass

New file `tests/test_readiness_hardening.py` covers:
- `probe_rate_limit`'s budget/window-reset/per-source-IP-isolation/testing-mode-bypass/stale-key-pruning behavior, in isolation against a minimal Flask app.
- `/ready` actually rejecting a source once its budget is spent, and confirming a rejected request never calls `ping()` at all.
- `get_pool_stats()`'s zero-before-any-connection case, a populated-pool case, the zero-division guard, and that its output never contains a DSN, host, or credential substring.
- `/metrics` rendering the three new gauges with correct values, never leaking the DSN into the scrape body, and surviving a broken stats provider without failing the rest of the scrape.

Verified: full backend suite (796 passed, 3 skipped — pre-existing, unrelated), plus `tests/test_observability.py`'s real PostgreSQL-backed readiness-leak test run against a local Postgres instance to confirm the new decorator doesn't disturb #306's fix. `ruff check .` and `ruff format --check .` clean.

## Related issue
Closes #296

## Checklist
- [ ] Matching CLI playbook / compliance framework mappings — not applicable, no scanner rule added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,12 +25,14 @@ OpenShield uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - Harvest Now Decrypt Later exposure window calculation per cryptographic asset
 - Deterministic Render deployment workflow for separate API and worker services
 - Terraform configuration for Render, Vercel, and GitHub OIDC
+- Database connection pool utilization and exhaustion telemetry (`openshield_db_pool_connections_*`) on `/metrics`
 
 ### Fixed
 
 - High-severity CodeQL findings in Python and JavaScript code
 - Security findings identified during Semgrep analysis
 - Sensitive identity metadata removed from scanner debug logging
+- `/ready` and `/metrics` rate-limited per source IP so the unauthenticated probe/scrape surface can no longer be used to exhaust the database connection pool
 
 ### Security
 
--- a/api/app.py
+++ b/api/app.py
@@ -126,7 +126,21 @@ def create_app() -> Flask:
     # Trust exactly one reverse-proxy hop (Render's edge) for the client IP
     # and scheme, so request.remote_addr reflects the real caller instead of
     # collapsing every client onto Render's proxy address. Rate limiting and
-    # any other per-IP logic depend on this being accurate.
+    # any other per-IP logic (api.observability.probe_rate_limit,
+    # api.rate_limit.rate_limit) depend on this being accurate.
+    #
+    # This is a trust boundary, not just a convenience setting: x_for=1 makes
+    # Flask take the *last* entry of an inbound X-Forwarded-For header as the
+    # real client IP, on the assumption that Render's edge is the only thing
+    # capable of appending to it before the request reaches this process. If
+    # the origin were ever reachable directly - bypassing Render's edge, e.g.
+    # a misconfigured DNS record or a leaked origin IP - a direct caller's own
+    # X-Forwarded-For header would be trusted as-is, and they could set it to
+    # a fresh IP on every request. Every per-IP control in this file (the
+    # probe-endpoint limiter, the Postgres-backed rate limiter) would then
+    # bucket each request as a "new" caller, which is equivalent to no rate
+    # limiting for that path at all. Keeping the origin unreachable except
+    # through Render's edge is what this setting's correctness depends on.
     app.wsgi_app = ProxyFix(app.wsgi_app, x_for=1, x_proto=1)
 
     # ------------------------------------------------------------------ #
--- a/api/app.py
+++ b/api/app.py
@@ -10,8 +10,15 @@ from flask import Flask, g, jsonify, request
 from flask_cors import CORS
 from werkzeug.middleware.proxy_fix import ProxyFix
 
-from api.models.finding import DatabaseManager
-from api.observability import configure_logging, get_request_id, init_app, init_sentry
+from api.models.finding import DatabaseManager, get_pool_stats
+from api.observability import (
+    configure_logging,
+    get_request_id,
+    init_app,
+    init_sentry,
+    probe_rate_limit,
+    set_pool_stats_provider,
+)
 
 load_dotenv()
 
@@ -41,6 +48,14 @@ _GENERATE_CMD = 'python -c "import secrets; print(secrets.token_urlsafe(32))"'
 # not touch write authorization.
 _KNOWN_ROLES = {"viewer", "operator", "admin"}
 _WRITE_ROLES = {"operator", "admin"}
+
+# Generous enough for legitimate manual or automated readiness checks from
+# one source, but bounded well under the default pool size
+# (DB_POOL_MAX_CONN=10) so a single caller can never claim more than half
+# the pool's capacity by itself, even if every allowed request in the
+# window lands at once. See probe_rate_limit's docstring for why this is
+# in-memory rather than the shared Postgres-backed rate_limit().
+_READY_MAX_REQUESTS_PER_WINDOW = 5
 
 
 def _is_production() -> bool:
@@ -121,6 +136,7 @@ def create_app() -> Flask:
     # every later before_request handler (including JWT auth) and to the
     # error handlers. Also mounts the public /metrics endpoint.
     init_app(app)
+    set_pool_stats_provider(get_pool_stats)
 
     # ------------------------------------------------------------------ #
     # Configuration & Security                                             #
@@ -269,6 +285,7 @@ def create_app() -> Flask:
         return jsonify({"status": "ok"})
 
     @app.get("/ready")
+    @probe_rate_limit(_READY_MAX_REQUESTS_PER_WINDOW)
     def ready():
         """Readiness probe: 200 when the database is reachable, else 503."""
         try:
--- a/api/models/finding.py
+++ b/api/models/finding.py
@@ -46,6 +46,44 @@ def _get_pool(dsn: str) -> "psycopg2.pool.ThreadedConnectionPool":
         return pool
 
 
+def get_pool_stats(dsn: Optional[str] = None) -> Dict[str, Any]:
+    """Return a point-in-time snapshot of the shared connection pool's utilization.
+
+    Reports only counts (in-use / idle / configured maximum) - never the DSN,
+    host, credentials, or anything else that could describe the database
+    deployment. Safe to expose on a public surface (Prometheus /metrics, the
+    /ready probe) precisely because it carries no operational secrets, only
+    capacity numbers an operator needs to see the pool approaching exhaustion
+    before it happens.
+
+    Returns zeroed stats with no pool created yet (no request has connected
+    since process start) rather than raising, so callers on the request path
+    (like /ready) never fail because of a stats lookup.
+    """
+    dsn = dsn or os.environ.get("DATABASE_URL", "")
+    with _POOLS_LOCK:
+        pool = _POOLS.get(dsn)
+
+    if pool is None:
+        return {"max_connections": _POOL_MAX_CONN, "in_use": 0, "idle": 0, "utilization_percent": 0.0}
+
+    # psycopg2's pool has no public stats API; _used/_pool/maxconn are the
+    # same attributes getconn()/putconn() themselves mutate under this same
+    # lock, so a snapshot taken while holding it can't land mid-mutation.
+    with pool._lock:
+        in_use = len(pool._used)
+        idle = len(pool._pool)
+        max_connections = pool.maxconn
+
+    utilization_percent = round((in_use / max_connections) * 100, 1) if max_connections else 0.0
+    return {
+        "max_connections": max_connections,
+        "in_use": in_use,
+        "idle": idle,
+        "utilization_percent": utilization_percent,
+    }
+
+
 FRAMEWORK_FILE_MAP = {
     "cis": "cis_azure_benchmark.json",
     "nist": "nist_csf.json",
--- a/api/observability.py
+++ b/api/observability.py
@@ -242,28 +242,52 @@ _PROBE_MAX_TRACKED_KEYS = 10_000
 # of traffic shape or how the sweep is scheduled.
 _PROBE_SWEEP_INTERVAL = 200
 _probe_lock = threading.Lock()
-# An OrderedDict, not a plain dict: every touch (new hit or a sweep pruning
-# it survives) moves a key to the end, so the front is always the least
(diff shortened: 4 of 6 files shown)


m-khan-97 (changes requested):
Parth, the overall direction is sound: keeping probe limiting out of PostgreSQL is the right boundary, rejected `/ready` calls do avoid the database, and the pool metrics expose counts rather than connection details. I also checked the current CI head; all 20 checks are green.

There is one blocking denial-of-service gap in the limiter itself. `_probe_hits` only removes an expired entry when that exact `(remote_addr, path)` key is requested again. A caller that continuously rotates source addresses—or spoofed forwarded addresses anywhere the trusted-proxy boundary is misconfigured—creates a new dictionary entry per address, and those one-shot keys are never revisited or removed. The new unauthenticated protection can therefore become an unbounded process-memory sink.

Please add bounded global cleanup rather than only same-key cleanup: for example, periodically sweep expired deques under the lock and enforce a hard maximum number of tracked keys with deterministic eviction. Add a regression test that inserts many distinct one-shot addresses, advances past the window, triggers cleanup through a different address, and proves stale keys are removed and the map remains capped. A `Retry-After` header on 429 would also make the endpoint friendlier to legitimate probes, but I do not consider that blocking.

Once the state is globally bounded, I will re-review promptly.

parthrohit22:
@m-khan-97 Fixed, in two parts:

1. **The gap you flagged.** `_probe_hits`'s per-call cleanup only ever pruned the one key the current request touched. Added the periodic global sweep and hard cap you asked for: a sweep every `_PROBE_SWEEP_INTERVAL` calls (not every call — a full-dict scan per request would defeat the point of a cheap in-memory limiter) that prunes every key whose hits have all expired, plus a hard `_PROBE_MAX_TRACKED_KEYS` ceiling with deterministic least-recently-touched eviction via an `OrderedDict` (every hit, including one that just survives a sweep, moves its key to the end). Also added the `Retry-After` header you suggested as non-blocking.
2. **Separately:** a `dev` merge onto this branch (picking up #294/#320's own changes to `api/app.py`) dropped `_READY_MAX_REQUESTS_PER_WINDOW`'s definition while keeping its usage on `/ready`, which broke CI (ruff's F821, and Backend Tests for the same reason). Restored it.

New tests cover exactly the scenario you described: many distinct one-shot addresses, advance past the window, trigger cleanup through different addresses, and prove the stale keys are gone and the map stays bounded — plus direct hard-cap/LRU-eviction tests and the `Retry-After` header.

Verified: full backend suite (862 passed, 5 skipped — pre-existing/environment-only), all 17 tests in the readiness-hardening file including the 5 new ones, ruff clean. All 20 CI checks are green on the current head. Ready for another look whenever you have a chance.

m-khan-97 (dismissed):
Parth, I verified the current head rather than relying on the update summary. The blocker is closed: stale one-shot source keys are now swept globally, tracked state has a hard 10,000-key ceiling, eviction is deterministic and activity-aware, and the rejected response includes `Retry-After`. The regression suite covers rotating addresses, global expiry cleanup, hard-cap behavior, LRU touch ordering, and the response header. The merge with #320 also restored the readiness budget constant correctly, and all 20 checks pass. Approving.

m-khan-97:
@ritiksah141 @TFT444, I completed the blocker rereview and approved `5854316`: the limiter now has global stale-key cleanup, a hard state cap, deterministic LRU eviction, and regression coverage. All 20 checks pass. Please provide the required independent review so it can proceed under the two-person rule.

TFT444 (dismissed):
**BLOCKER — LRU eviction bypass: `move_to_end(key)` is called unconditionally on every request including rejected ones, keeping an attacker's slot permanently warm while legitimate monitoring IPs that go quiet drift toward eviction.** Once a legitimate IP's entry is evicted its hit history resets and it gets a fresh budget; the attacker's key never ages out — directly defeating the eviction-based memory cap; guard `move_to_end(key)` with `if allowed`.

**BLOCKER — Shared sweep uses wrong window: `_probe_hits` is a module-level dict shared across all decorated endpoints, but the sweep passes the triggering endpoint's `window_seconds` closure to `_sweep_expired_probe_hits`, applying it to every key including entries owned by endpoints with different windows.** Both endpoints currently default to 10 s so the bug is dormant, but the first non-default-window endpoint added will cause premature budget resets or lingering stale entries across all endpoints — store `window_seconds` per key so the sweep applies the correct cutoff per entry.

**MAJOR — `ProxyFix(x_for=1)` makes `remote_addr` fully attacker-controlled if the origin is reachable directly** (Render origin IPs surface in CT logs and Shodan); an attacker sending arbitrary `X-Forwarded-For` headers gets a fresh rate-limit bucket on every request, fully neutralising per-IP enforcement — add an explicit code comment stating the trusted-proxy assumption and what fails when it is violated.

parthrohit22:
@TFT444 All 3 fixed:

1. `move_to_end(key)` now only runs on a request that actually counts against the budget — a rejected request no longer keeps an over-budget key artificially warm at the attacker's expense of quiet legitimate keys.
2. Added `_ProbeEntry` to carry `window_seconds` per tracked key instead of using whichever endpoint's request triggered the sweep — a sweep from one endpoint can no longer misapply its window to another endpoint's keys.
3. Expanded the comment on `ProxyFix(x_for=1)` in `api/app.py` to spell out the trust assumption and exactly what breaks (every per-IP control becomes a no-op) if the origin is ever reachable outside Render's edge.

New regression tests for both behavioral fixes — one proves a hammered exhausted key gets evicted correctly instead of staying warm, the other runs two endpoints with different windows sharing the tracked-key dict and proves a sweep from one doesn't clobber the other's entries.

Full suite green (864 passed), ruff clean, CI running now.

parthrohit22:
## Response to review

Rebased onto latest `dev` (was 1 commit behind #279; one trivial conflict in `api/app.py` where #320's `_KNOWN_ROLES`/`_WRITE_ROLES` and this branch's `_READY_MAX_REQUESTS_PER_WINDOW` were added at the same spot — both kept). All three findings from @TFT444's review are addressed in `api/observability.py`; @m-khan-97's memory-bound blocker was closed earlier in the same commit.

**BLOCKER — LRU eviction bypass (`move_to_end` on rejected requests)** — fixed. `_probe_hits.move_to_end(key)` now runs only inside the `if allowed:` branch, after appending the hit. A rejected (over-budget) request no longer refreshes the key's eviction position, so an abusive key that never lets its own budget recover drifts to the front and is evicted before quiet legitimate keys.
Regression test: `test_probe_rate_limit_rejected_requests_do_not_refresh_eviction_order` — one key hammered past its budget while two others stay quiet; at the hard cap, a new key evicts the *abuser*, not the quiet legitimate keys. (`test_probe_rate_limit_touching_a_key_protects_it_from_eviction` covers the allowed-request side.)

**BLOCKER — shared sweep uses the wrong window** — fixed. `window_seconds` is now stored per key on a new `_ProbeEntry` (`__slots__ = ("hits", "window_seconds")`). Both `_sweep_expired_probe_hits()` and the per-call prune in `probe_rate_limit` compute `cutoff = now - entry.window_seconds`, so a sweep triggered by one endpoint applies each entry's own endpoint window rather than the triggering closure's.
Regression test: `test_probe_rate_limit_sweep_uses_each_keys_own_window_not_the_callers` — a 1 s-window endpoint and a 1000 s-window endpoint share `_probe_hits`; time advances past the short window only; a request to the long-window endpoint triggers the sweep; the short-window key is evicted and the long-window key survives.

**MAJOR — `ProxyFix(x_for=1)` trust assumption** — documented. `api/app.py` now carries a comment at the `ProxyFix` config spelling out that `x_for=1` is a trust boundary (only Render's edge may append to `X-Forwarded-For`), and that if the origin is ever reachable directly a caller can forge a fresh IP per request, making every per-IP control here (the probe limiter and the Postgres-backed `rate_limit()`) trivially bypassable.

**Non-blocking — `Retry-After` on 429** — added. The rejection returns `429` with `Retry-After: <window_seconds>`. Covered by `test_probe_rate_limit_429_includes_retry_after_header`.

`tests/test_readiness_hardening.py`: 19 passed. `ruff check .` / `ruff format --check .`: clean. Full backend suite: 953 passed, 3 skipped (2 unrelated `test_observability.py` failures locally need the CI Postgres service).

@TFT444 @m-khan-97 — re-review please.

TFT444 (approved):
All three findings from the previous CHANGES_REQUESTED are addressed. Verified against the diff at head commit 8241d47.

**Blocker #1 (LRU eviction bypass) — FIXED**

`move_to_end(key)` is now inside `if allowed:` at `api/observability.py` lines 356-365. A rejected request no longer refreshes its key's position in the eviction order. A regression test (`test_probe_rate_limit_rejected_requests_do_not_refresh_eviction_order`) proves the exact scenario: an attacker who exhausts their one-request budget and keeps hammering has their key evicted in favour of a quiet legitimate address, not the other way around.

**Blocker #2 (Shared sweep uses wrong window) — FIXED**

`window_seconds` is now stored per tracked entry via the `_ProbeEntry` dataclass (`__slots__ = ("hits", "window_seconds")`). `_sweep_expired_probe_hits` reads `entry.window_seconds` for each key rather than taking the triggering request's closure value. The `_ProbeEntry` docstring explicitly names the bug and why both endpoints sharing a 10 s default kept it dormant. A regression test (`test_probe_rate_limit_sweep_uses_each_keys_own_window_not_the_callers`) covers two endpoints with different windows (1 s and 1000 s), advances past only the short one, triggers a sweep via the long-window endpoint, and confirms the short-window key is evicted while the long-window key survives.

**Major (ProxyFix trusted-proxy comment) — FIXED**

`api/app.py` lines 64-76 add a comment that names the threat (attacker sends arbitrary `X-Forwarded-For`, gets a fresh rate-limit bucket per request), identifies what the setting's correctness depends on (origin unreachable except through Render's edge), and explains what breaks when that assumption is violated. This is the right level of detail for a trust boundary. `docs/deployment/render.md` also adds an operational note pointing to the same assumption for anyone deploying behind a different proxy.

**Additional observations (non-blocking)**

- The periodic sweep (`_PROBE_SWEEP_INTERVAL = 200`) and hard cap (`_PROBE_MAX_TRACKED_KEYS = 10_000`) address the unbounded-memory concern m-khan-97 raised in the earlier round. Both have dedicated regression tests.
- `_probe_call_count` is an unbounded incrementing int but Python integers do not overflow, so this is fine in practice.
- `get_pool_stats()` reads `pool._used`, `pool._pool`, and `pool.maxconn` under `pool._lock`, which is the same lock psycopg2's `ThreadedConnectionPool` uses internally for `getconn`/`putconn`, so the snapshot is consistent. The test covers the zero-division guard and confirms no DSN/host/credential appears in the output.
- 796 tests passing, CI green per the PR description. Test isolation is correctly handled by the `autouse` fixture that clears `_probe_hits` before and after every test.

Approving.

ritiksah141 (approved):

PR #320, opened 2026-08-28
fix(api): enforce token expiry, role, and subscription authorization

parthrohit22:
## What does this PR do?
Partially addresses #294: enforces token expiry and role on every accepted JWT, gates every write endpoint to `operator`/`admin` roles, and adds an optional subscription authorization allowlist on `POST /api/scans/trigger`. This is the immediate, in-repo, dependency-free containment slice of #294 - see the scope note below for what's deliberately not in this PR.

## Type of change
- [x] Bug fix
- [x] API endpoint
- [x] Documentation

## Scope note

#294 is really two bodies of work of very different size and dependency profile. This PR is the first only; the second is flagged, not forced.

**In this PR** - confirmed live before starting (`verify_jwt()` only checked the signature and expiry-if-present, `role` was decoded into `g.user` but never read by any route, and the baseline schema has no tenant/org column at all):

1. **No-expiry tokens.** `jwt.decode()` now requires `exp` (`options={"require": ["exp"]}`). A token missing it is rejected the same as an invalid signature — closes the specific hole where the documented `VITE_JWT_TOKEN` demo credential "works without rotation" because it never expires.
2. **Role never enforced.** `verify_jwt()` now requires a recognized `role` (`viewer`/`operator`/`admin`) on every accepted token, and rejects any non-`GET`/`HEAD` request from a `viewer` token with `403`. Enforced once in the middleware, so it covers every current write route (scan trigger, all 5 AI endpoints) and any future one uniformly, rather than needing a per-route check. Confirmed this holds under `OPENSHIELD_PUBLIC_DEMO=true` too — demo mode only ever widens *read* access, never write.
3. **No subscription authorization.** `POST /api/scans/trigger` accepted an arbitrary caller-supplied `subscription_id` with no check that the caller was entitled to it — any `operator`/`admin` token could trigger a scan against any subscription. Added an optional `OPENSHIELD_AUTHORIZED_SUBSCRIPTIONS` comma-separated allowlist, which is the "configured subscription allowlist" this issue itself names as an acceptable single-tenant interim. Left unset, behavior is unchanged from today (no deployment is silently broken by this PR); the API now logs a loud startup warning when it's left unset, matching the existing `ALLOWED_ORIGINS`/`JWT_SECRET` convention in the same function.
4. **`scripts/generate_demo_jwt.py`** minted a permanent, no-expiry token by design. Now requires an expiry (`DEMO_JWT_TTL_HOURS`, default 24h); docstring/output updated to stop claiming it works without rotation.
5. `docs/api-reference.md` and `.env.example` document all of the above.

**Deliberately not in this PR**, flagged rather than forced: the durable-boundary body (Entra/OIDC Authorization Code + PKCE, JWKS validation, per-tenant schema/RLS, cross-tenant integration tests) needs an actual identity-provider app registration (client ID/secret, redirect URIs) against this deployment's real domains, plus a product decision on which IdP and whether multi-tenancy is in scope for a first release. Both are outside what a PR to this repository can provision or decide alone. #294 stays open for that.

Also deliberately untouched: `frontend/src/App.jsx`'s local-dev `dev-local-token` bootstrap. It is not a valid signed JWT (`verify_jwt()` rejects it outright outside demo mode) and only ever worked because `OPENSHIELD_PUBLIC_DEMO` bypasses the JWT check for `GET` — removing it without a replacement login flow would break local dev for every contributor while fixing nothing real, since it was never a working credential. The actual leaked-credential risk is `VITE_JWT_TOKEN`, which is what items 1-2 above bound.

## Testing
- [x] All CI checks pass

New/updated coverage:
- `tests/test_auth.py`: missing-`exp` rejected, missing/unrecognized-`role` rejected, `viewer` can read, `viewer` cannot write (including under demo mode), `operator`/`admin` can write. All pre-existing cases in this file (and `tests/test_async_scan_persistence.py`, `smoke_test.py`) already used tokens carrying `role`+`exp`, so nothing broke.
- `tests/test_subscription_authorization.py` (new): unset allowlist preserves today's behavior, allowlisted/non-allowlisted subscriptions, case-insensitive comparison, blank-entry handling, startup-warning presence/absence.

Verified: full backend suite (796 passed, 5 skipped — pre-existing/environment-only, confirmed identical skip count against unmodified `dev`). `ruff check .` and `ruff format --check .` clean.

## Related issue
Partially addresses #294 — the durable multi-tenant/OIDC boundary remains open, see scope note above.

## Checklist
- [ ] Matching CLI playbook / compliance framework mappings — not applicable, no scanner rule added

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,6 +4,11 @@ DATABASE_URL=postgresql://openshield:openshield@localhost:5432/openshield
 # Auth
 JWT_SECRET=change-me-in-production
 
+# Optional - comma-separated subscription_id allowlist for POST /api/scans/trigger.
+# Unset accepts any subscription_id (matches historical behavior); the API
+# logs a startup warning when this is unset. See docs/api-reference.md.
+OPENSHIELD_AUTHORIZED_SUBSCRIPTIONS=
+
 # AI providers - add at least one
 ANTHROPIC_API_KEY=
 GROQ_API_KEY=
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ OpenShield uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - AI provider errors no longer expose upstream response details
 - Request body limits, AI rate limiting, and playbook path validation added
 - GitHub Actions dependencies pinned to immutable commit SHAs
+- JWTs must now carry an `exp` claim and a recognized `role`; a `viewer` token can no longer perform any write operation (scan trigger, AI endpoints), and `POST /api/scans/trigger` now checks `subscription_id` against an optional `OPENSHIELD_AUTHORIZED_SUBSCRIPTIONS` allowlist
 
 ## [0.3.0] - 2026-07-08
 
--- a/api/app.py
+++ b/api/app.py
@@ -32,6 +32,16 @@ _MIN_JWT_SECRET_LENGTH = 32
 _MAX_AUTHORIZATION_HEADER_LENGTH = 8192
 _GENERATE_CMD = 'python -c "import secrets; print(secrets.token_urlsafe(32))"'
 
+# A token's signature proves who signed it, not what the bearer is allowed to
+# do. Every accepted token must carry one of these roles (see issue #294):
+# a missing/unrecognized role is treated the same as an invalid signature.
+# Only operator/admin may perform a write (any non-GET/HEAD); viewer is
+# read-only. This is enforced regardless of demo mode - public_demo only
+# ever widens *read* access to skip the token requirement entirely, it does
+# not touch write authorization.
+_KNOWN_ROLES = {"viewer", "operator", "admin"}
+_WRITE_ROLES = {"operator", "admin"}
+
 
 def _is_production() -> bool:
     return (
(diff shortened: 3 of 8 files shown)


ritiksah141 (approved):
Approved. Verified end to end: exp is now required (MissingRequiredClaimError is handled by the existing InvalidTokenError path), role enforcement in the before_request middleware covers all 7 write endpoints (scans trigger, scans enrich, 5 AI endpoints) uniformly including any future route, the write gate holds even with OPENSHIELD_PUBLIC_DEMO=true, and the allowlist check runs after uuid_string normalization and before any DB write, so there is no case or format bypass. Confirmed /api/scans/<scan_id>/enrich takes no subscription input and ai.py has no subscription_id anywhere, so gating the trigger endpoint alone is correct. Unset allowlist preserves historical behavior with a loud startup warning. Backend suite green locally, ruff clean, CI green.

Minor nits, none blocking:

1. docs/validation/FRONTEND_API_TESTING.md:67 still describes VITE_JWT_TOKEN as just a "Pre-signed JWT for dev" with no mention of the new exp/role requirements. api-reference.md was updated but this file was missed.
2. Deployment note: the frontend calls write endpoints (triggerScan in utils/api.js, all of aiApi.js), so with the viewer-only demo token those calls will 403 in the deployed demo. That is the intended containment here, but the buttons will be dead until the frontend removes or gates them, which is part of the remaining #294 scope.
3. Worth one line in the docs: if the allowlist is set but the AZURE_SUBSCRIPTION_ID env fallback is not on it, empty-body triggers will 403. Correct fail-closed behavior, but operators need to know their own subscription must be on the list.
4. Cosmetic in tests/test_subscription_authorization.py: record.message vs record.getMessage() inconsistency, and import logging inside the test functions.

TFT444 (approved):
Verified: exp enforcement, role gate on all 7 write endpoints, and subscription allowlist all correct.

PR #293, opened 2026-08-20
Add enterprise AKS and workload security controls

ritiksah141:
## Summary

Implements the complete enterprise AKS and container workload security backlog from issue #255 with fifteen evidence-rich controls covering API exposure, Kubernetes network policy, Defender for Containers, secret protection, privileged and host access, cluster administration, and image supply chain policy.

## What changed

- Added `AZ-AKS-007` to require a private API server or organization-approved authorized IP ranges.
- Added `AZ-AKS-008` to require Azure, Calico, or Cilium network policy enforcement.
- Added `AZ-AKS-009` to detect eligible namespaces without a NetworkPolicy.
- Added `AZ-AKS-010` to require Microsoft Defender for Containers.
- Added `AZ-AKS-011` to detect native Kubernetes Secret references without Azure Key Vault KMS protection.
- Added `AZ-AKS-012` to require automatic rotation when the Key Vault Secrets Store CSI provider is enabled.
- Added `AZ-AKS-013` through `AZ-AKS-017` as separate controls for privileged containers, hostNetwork, hostPID, hostIPC, and hostPath volumes.
- Added `AZ-AKS-018` to detect cluster-admin bindings outside the approved subject allowlist.
- Added `AZ-AKS-019` through `AZ-AKS-021` as separate controls for untrusted registries, latest or implicit latest tags, and mutable non-digest image references.
- Added a cached AKS evidence collector for ARM, Defender for Cloud, Kubernetes namespaces, workloads, NetworkPolicies, and ClusterRoleBindings.
- Added strict policy loading through `OPENSHIELD_AKS_SECURITY_POLICY` and Kubernetes access through `OPENSHIELD_AKS_KUBECONFIG`.
- Added cluster, namespace, workload, container, image, subject, role, evidence, remediation, permissions, severity, confidence, and unknown reason metadata.
- Added an individual review-gated remediation playbook for every new rule.
- Added complete CIS, NIST CSF, ISO 27001, and SOC 2 mappings with explicit reviewed `N/A-*` CIS decisions where no direct benchmark mapping exists.
- Updated backend documentation, rule references, architecture guidance, contributor guidance, and the changelog without modifying `/website`.

## False-finding safeguards

- Empty AKS inventory is treated as not applicable.
- Missing policy, malformed policy, unavailable credentials, unauthorized Kubernetes access, and API failures are indeterminate and never create findings.
- Stopped clusters and unsupported provisioning states are reported as unknown rather than pass or fail.
[...]

PR #321, opened 2026-08-29
feat(engine): add rule evaluation coverage contract (#263)

dipeshrayg:
## What does this PR do?
Adds the rule evaluation coverage contract from #263: rules can now report a status for every resource they looked at (PASS included), so a scan that never ran, errored, or came from an unmigrated rule can no longer be silently read as a pass.

## Type of change
- [x] Bug fix
- [x] API endpoint
- [x] Documentation

## Testing
- [x] Full test suite passes locally (821 passed, 9 skipped: the 4 populated-database tests below are skipped without a live Postgres and run for real in CI)
- [x] All seven CI-equivalent checks reproduced locally and passing (rule structure, credential scan, playbook existence/syntax, compliance JSON validity, API syntax, compliance cross-reference, `alembic heads` resolves to exactly one head)

## Related issue
Addresses #263

## What changed
- `scanner/evaluation.py`: `EvaluationStatus` (PASS/FAIL/UNKNOWN/ERROR/NOT_APPLICABLE), `RuleEvaluation` dataclass, `subscription_scope_id()` for canonical non-empty scope identifiers, and `aggregate_status()` implementing the conservative FAIL > ERROR > UNKNOWN > PASS > NOT_APPLICABLE roll-up order.
- New migration `alembic/versions/3f59f83a5253_rule_evaluations.py`, chained onto #308's head (`d8e4f6a1b2c3`). `rule_evaluations` table with CHECK constraints on the five statuses, a non-empty `resource_id`, and a required `reason_code` for UNKNOWN/ERROR/NOT_APPLICABLE. `finding_id` is a nullable FK, populated only for FAIL evaluations.
- `scanner/engine.py`: calls `rule.evaluate()` when a rule exposes it. A rule without `evaluate()` gets one UNKNOWN/LEGACY_RULE_NOT_MIGRATED row instead of silently having no coverage. An `evaluate()` exception (or a non-list return) produces one ERROR at the canonical rule/subscription scope. A FAIL evaluation contributes its own finding only when `scan()` hasn't already reported the same `(rule_id, resource_id)`, so a rule implementing both never double-counts.
- `api/models/finding.py`: `save_scan()` persists `rule_evaluations` in the same transaction as findings and backfills `finding_id` for FAIL rows via a join on `(scan_id, rule_id, resource_id)`. `get_compliance_score()` now derives each control's status from `rule_evaluations` (never from finding absence), returns `evaluated`/`passed`/`failed`/`unknown`/`error`/`not_applicable` counts separately, and excludes UNKNOWN/ERROR from the numerator so they can never improve the score. Response carries `contract_version: "2"`.
- `scanner/rules/az_kv_006.py`: migrated as the one reference rule. `evaluate()` reports PASS/FAIL per vault, UNKNOWN for a vault missing its properties payload (previously silently skipped by `scan()`), and NOT_APPLICABLE when the vault list is empty (since `AzureClient.get_key_vaults()` can't distinguish "no vaults" from "the list call failed"; closing that gap is the deferred ARG cross-check).
- `docs/adding-a-rule.md`: short opt-in section documenting the `evaluate()` contract for future rule migrations.

## Regression coverage
- Legacy rule (no `evaluate()`) is recorded as UNKNOWN/LEGACY_RULE_NOT_MIGRATED, never PASS.
- `evaluate()` exception produces ERROR at canonical scope instead of vanishing.
- `evaluate()` returning a non-list is treated as an error, not silently accepted.
- A FAIL evaluation contributes its own finding when `scan()` didn't already report it.
- A FAIL evaluation does not duplicate a finding `scan()` already reported.
- `rule_evaluations` rows persist in the same transaction as findings, with `finding_id` correctly linked (FAIL) or left null (UNKNOWN/ERROR/PASS/NOT_APPLICABLE).
- A retried scan replaces prior `rule_evaluations` rows atomically, same as findings.
- `aggregate_status()` conservative ordering (FAIL beats ERROR beats UNKNOWN beats PASS beats NOT_APPLICABLE) under every input order.
- `get_compliance_score()`: a clean scan with real PASS evaluation rows shows PASS; a clean scan with zero evaluation rows shows UNKNOWN, not PASS (the bug this PR fixes); a remediated rule with a new PASS row shows PASS; worst-severity aggregation across findings is unchanged.
- AZ-KV-006 `evaluate()`: compliant vault -> PASS, non-compliant -> FAIL with the finding embedded, missing properties -> UNKNOWN, empty vault list -> NOT_APPLICABLE, mixed vaults report one status each.
- CHECK constraints (populated-database, run against CI's Postgres): reject an unsupported status, reject an empty `resource_id`, reject a NULL `reason_code` on UNKNOWN, and accept a valid PASS row with `finding_id` left null.
- Single Alembic head and correct chain onto `d8e4f6a1b2c3`.

## Scope
Matches what was agreed on #263: the contract, persistence/aggregation, and one end-to-end reference rule (AZ-KV-006). The Azure Resource Graph completeness cross-check stays a separate follow-up, as discussed.

## Checklist

### DIFF
--- a/alembic/versions/3f59f83a5253_rule_evaluations.py
+++ b/alembic/versions/3f59f83a5253_rule_evaluations.py
@@ -1,0 +1,85 @@
+"""Add rule_evaluations: per-resource coverage, not just findings (#263).
+
+Revision ID: 3f59f83a5253
+Revises: d8e4f6a1b2c3
+Create Date: 2026-08-29 00:00:00.000000
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# Revision identifiers, used by Alembic.
+revision: str = "3f59f83a5253"
+down_revision: Union[str, Sequence[str], None] = "d8e4f6a1b2c3"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+_STATUS_CONSTRAINT = "ck_rule_evaluations_status_v1"
+_SCOPE_CONSTRAINT = "ck_rule_evaluations_resource_id_not_empty"
+_REASON_CONSTRAINT = "ck_rule_evaluations_reason_code_required"
+
+
+def upgrade() -> None:
+    op.create_table(
+        "rule_evaluations",
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column("scan_id", postgresql.UUID(), nullable=False),
+        sa.Column("rule_id", sa.Text(), nullable=False),
+        sa.Column("resource_id", sa.Text(), nullable=False),
+        sa.Column("resource_type", sa.Text(), nullable=False, server_default=sa.text("''")),
+        sa.Column("status", sa.Text(), nullable=False),
+        sa.Column("reason_code", sa.Text(), nullable=True),
+        sa.Column("reason", sa.Text(), nullable=True),
+        sa.Column("evidence", postgresql.JSONB(), server_default=sa.text("'{}'::jsonb"), nullable=True),
+        # Nullable: only set for FAIL evaluations, and only once the finding
+        # row exists. Populated in the same transaction as the finding insert
+        # (see DatabaseManager.save_scan), never inferred after the fact.
+        sa.Column("finding_id", sa.Integer(), nullable=True),
+        sa.Column("evaluated_at", sa.DateTime(timezone=True), nullable=False),
+        sa.ForeignKeyConstraint(["scan_id"], ["scans.scan_id"], name="rule_evaluations_scan_id_fkey"),
+        sa.ForeignKeyConstraint(
+            ["finding_id"], ["findings.id"], name="rule_evaluations_finding_id_fkey", ondelete="SET NULL"
+        ),
+        sa.PrimaryKeyConstraint("id", name="rule_evaluations_pkey"),
+        # One coverage statement per rule per resource per scan. Also gives
+        # the persistence-layer FAIL -> finding_id backfill a stable join key.
+        sa.UniqueConstraint("scan_id", "rule_id", "resource_id", name="uq_rule_evaluations_scan_rule_resource"),
+    )
+
+    op.create_index("idx_rule_evaluations_scan_id", "rule_evaluations", ["scan_id"], unique=False)
+    op.create_index("idx_rule_evaluations_rule_id", "rule_evaluations", ["rule_id"], unique=False)
+    op.create_index("idx_rule_evaluations_status", "rule_evaluations", ["status"], unique=False)
+
+    op.create_check_constraint(
+        _STATUS_CONSTRAINT,
+        "rule_evaluations",
+        "status IN ('PASS', 'FAIL', 'UNKNOWN', 'ERROR', 'NOT_APPLICABLE')",
+    )
+    # A canonical scope identifier is required — never an empty string standing
+    # in for "no specific resource" (that collides across rules/subscriptions).
+    op.create_check_constraint(
+        _SCOPE_CONSTRAINT,
+        "rule_evaluations",
+        "resource_id <> ''",
+    )
+    # UNKNOWN/ERROR/NOT_APPLICABLE must always explain themselves; only PASS
+    # and FAIL are self-evident from the status alone.
+    op.create_check_constraint(
+        _REASON_CONSTRAINT,
+        "rule_evaluations",
+        "status NOT IN ('UNKNOWN', 'ERROR', 'NOT_APPLICABLE') "
+        "OR (reason_code IS NOT NULL AND reason_code <> '')",
+    )
+
+
+def downgrade() -> None:
+    op.drop_constraint(_REASON_CONSTRAINT, "rule_evaluations", type_="check")
+    op.drop_constraint(_SCOPE_CONSTRAINT, "rule_evaluations", type_="check")
+    op.drop_constraint(_STATUS_CONSTRAINT, "rule_evaluations", type_="check")
+    op.drop_index("idx_rule_evaluations_status", table_name="rule_evaluations")
+    op.drop_index("idx_rule_evaluations_rule_id", table_name="rule_evaluations")
+    op.drop_index("idx_rule_evaluations_scan_id", table_name="rule_evaluations")
+    op.drop_table("rule_evaluations")
--- a/alembic/versions/3f59f83a5253_rule_evaluations.py
+++ b/alembic/versions/3f59f83a5253_rule_evaluations.py
@@ -70,8 +70,7 @@ def upgrade() -> None:
     op.create_check_constraint(
         _REASON_CONSTRAINT,
         "rule_evaluations",
-        "status NOT IN ('UNKNOWN', 'ERROR', 'NOT_APPLICABLE') "
-        "OR (reason_code IS NOT NULL AND reason_code <> '')",
+        "status NOT IN ('UNKNOWN', 'ERROR', 'NOT_APPLICABLE') OR (reason_code IS NOT NULL AND reason_code <> '')",
     )
 
 
--- a/api/models/finding.py
+++ b/api/models/finding.py
@@ -214,7 +214,6 @@ class DatabaseManager:
                 # keeps the scan header, child rows, and recomputed score in
                 # agreement instead of duplicating findings on every attempt.
                 cur.execute("DELETE FROM findings WHERE scan_id = %s", (scan_result["scan_id"],))
-                cur.execute("DELETE FROM rule_evaluations WHERE scan_id = %s", (scan_result["scan_id"],))
                 finding_id_by_key: Dict[Any, int] = {}
                 for f in findings:
                     cur.execute(
@@ -257,8 +256,14 @@ class DatabaseManager:
                 # is durably linked to the finding row it corresponds to
                 # right here, in the same transaction, instead of leaving
                 # callers to infer the relationship from rule_id/resource_id.
+                #
+                # Upserted rather than replaced wholesale: a retried/replayed
+                # scan result must converge on the same rows instead of a
+                # delete-then-reinsert racing a concurrent reader that could
+                # briefly see zero coverage for a scan that already has some.
                 evaluated_at = completed_at
-                for evaluation in scan_result.get("evaluations", []):
+                evaluations = scan_result.get("evaluations", [])
+                for evaluation in evaluations:
                     status = evaluation.get("status")
                     finding_id = None
                     if status == EvaluationStatus.FAIL:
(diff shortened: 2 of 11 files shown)


m-khan-97:
@dipeshrayg, this remains the canonical review track for #263’s evaluation semantics: status vocabulary, per-resource evidence rows, reason-code constraints, finding linkage, aggregation, and the reference-rule migration. Please keep it draft while we align its migration and persistence API with #325. Do not widen it into leases, admission, enrichment, or finding lifecycle; those belong to separate layers.

dipeshrayg:
Understood, keeping this in draft and staying within the evaluation-semantics scope you outlined. No leases, admission, enrichment, or finding lifecycle here.

I looked at #325 to see where the overlap is concretely: its `f2b6d8e1a4c9` migration also creates `rule_evaluations`, chained onto the same parent (`d8e4f6a1b2c3`) this PR's `3f59f83a5253` chains onto. Same table, same parent revision, two different migrations, so as they stand neither can merge cleanly against the other (two Alembic heads plus a duplicate-table conflict). The schemas are close but not identical: #325's version doesn't carry the reason_code-required-for-UNKNOWN/ERROR/NOT_APPLICABLE CHECK constraint this PR added per your conditions above, and it upserts via `ON CONFLICT` for the lease/idempotency retry story rather than replacing rows per scan.

I'll hold off on touching the migration or persistence code here until you've decided how the two should line up. Happy to adjust either the migration chain or the persistence approach once that's settled, whichever way you want it to land.

ritiksah141:
Cross-PR coordination note: #325 (scan leases/fencing/idempotency, #303) also creates a rule_evaluations table and its own persistence semantics, which overlaps this PR. Per m-khan, this PR is the agreed #263 evaluation-contract implementation, so #325 should drop its copy and keep only the fencing/idempotency layer. Sharing the side-by-side here so the reconciliation is clear on both sides.

## Schema: rule_evaluations

| Aspect | #321 (this PR, 3f59f83a5253) | #325 (f2b6d8e1a4c9) |
|---|---|---|
| Core columns (id, scan_id, rule_id, resource_id, resource_type, status, reason_code, reason, evidence, finding_id, evaluated_at) | yes | identical |
| PK, FK scan_id, FK finding_id ON DELETE SET NULL | yes | identical |
| Unique (scan_id, rule_id, resource_id), same constraint name | yes | same name |
| Status CHECK (5 values), same constraint name | yes | same name |
| resource_id <> '' CHECK | yes | yes |
| Index on scan_id | yes | yes |
| Index on rule_id | yes | no |
| Index on status | yes | no |
| CHECK: reason_code required for UNKNOWN/ERROR/NOT_APPLICABLE | yes | no |

The core columns, keys, and constraints overlap and share names, so whichever migration runs second fails with "relation already exists." This PR is the strict superset: it adds the rule_id and status indexes and the reason_code-required CHECK.

## Producer (who emits evaluations)

| | #321 (this PR) | #325 |
|---|---|---|
| scanner/evaluation.py (EvaluationStatus, RuleEvaluation, subscription_scope_id, aggregate_status) | added | none |
| engine.run_scan calls evaluate() per rule and collects evaluations | added | none, engine untouched |
| FAIL evaluation contributes a finding (deduped vs scan() by rule_id+resource_id) | added | none |
| Returns evaluations in the scan result | added | reads it but never populated |

Only this PR makes evaluations observable; #325 has no producer, so its evaluations would always be empty in production.

## save_scan persistence semantics

Both PRs are full rewrites of the same method with incompatible idempotency models, so the conflict is larger than the table.

| | #321 (this PR) | #325 |
|---|---|---|
| Signature | save_scan(scan_result) | save_scan(scan_result, lease_owner, fencing_token) |
| Fencing / lease check | none | SELECT FOR UPDATE owner+token+unexpired, else LostLease |
| Findings idempotency | DELETE all, re-insert (full replace) | UPSERT ON CONFLICT (scan_id, finding_key) + delete-absent |
| Evaluations idempotency | DELETE all, plain re-insert | UPSERT ON CONFLICT (scan_id, rule_id, resource_id) + delete-absent |
| FAIL eval linked to finding via (rule_id, resource_id) in same txn | yes | yes |
| Requires rule_id + resource_id on every evaluation | no | yes, raises ValueError |

## Compliance score (the actual #263 bug)

| | #321 (this PR) | #325 |
|---|---|---|
| Rewrites get_compliance_score to read statuses from rule_evaluations | yes | no |
| aggregate_status FAIL > ERROR > UNKNOWN > PASS > NOT_APPLICABLE | yes | no |
| No evaluation row reports UNKNOWN instead of PASS | yes | no |
| Score excludes NOT_APPLICABLE from denominator, never counts UNKNOWN/ERROR as pass | yes | no |

Only this PR fixes the score-inflation bug.

## Recommended resolution

1. This PR owns the evaluation contract: schema, scanner/evaluation.py, engine producer, aggregation, and the compliance-score fix.
2. #325 drops its rule_evaluations table creation and its evaluation upsert, and keeps everything else: leases, fencing, admission, enrichment, metrics.
3. The real merge point is save_scan, not just the table. The final save_scan should keep #325's fenced, upsert skeleton (ownership check, lease clear, findings upsert by finding_key) and fold this PR's evaluation field set and FAIL-to-finding_id linkage into that same fenced transaction. Given the replay-safety goal of #303, evaluations should use the upsert-plus-delete-absent model.
4. Ordering: this only composes cleanly if this PR merges first (or both merge as a deliberate pair), then #325 rebases its remaining migrations on top of 3f59f83a5253. Both are currently OPEN and both branch off d8e4f6a1b2c3, so merging in the wrong order produces two Alembic heads and breaks the single-head CI gate.

One integration detail for whoever reconciles: this PR's engine adds evaluate()-derived FAIL findings to the findings list, and #325 derives finding_key from rule_id plus resource scope plus discriminator. Those compose, but make sure evaluate()-derived findings get stable finding_keys so the upsert stays idempotent.

dipeshrayg:
Thanks for the detailed side by side, that matches what I found and goes further. Agreed with the recommended resolution: this PR owns the schema, scanner/evaluation.py, the engine producer, aggregation, and the compliance-score fix, #325 keeps leases/fencing/admission/enrichment/metrics and drops its rule_evaluations copy.

On the save_scan reconciliation: agreed the real merge point is there, not just the table, and that the final version should be #325's fenced upsert skeleton with this PR's evaluation field set and FAIL-to-finding_id linkage folded in, using upsert-plus-delete-absent for evaluations to match #303's replay-safety model rather than this PR's current per-scan delete-and-reinsert. I'll hold off implementing that here until the ordering is confirmed, since it depends on #325's lease/fencing columns and finding_key concept that don't exist in this PR's schema.

Also agreed on the finding_key point for evaluate()-derived findings, once that reconciliation happens I'll make sure AZ-KV-006's FAIL findings produce a stable, discriminator-safe key so the upsert in #325's model stays idempotent.

Ready whenever the merge order is decided.

ritiksah141:
Ordering proposal for this PR and #325.

My suggestion: merge this PR (#321) first as the canonical evaluation contract, then rebase #325 on top of it. #325 would drop its duplicate rule_evaluations copy and re-chain its migrations on top of 3f59f83a5253, keeping its leases, fencing, admission, enrichment, and metrics work and folding this PR's evaluation writes into its fenced save_scan. This matches the resolution agreed above.

Two reasons I want to confirm the direction now. @SHAURYAKSHARMA24 does not seem active on #325 at the moment, and my own draft PR #293 also depends on this resolution landing, since it needs the canonical contract in place before it can move forward.

@m-khan-97 can you approve the ordering, #321 first? Once you confirm, @dipeshrayg this PR would be clear to make the save_scan change we discussed (upsert plus delete-absent for evaluations) and get ready to merge. Happy to help drive it however is useful.

dipeshrayg:
Pushed the save_scan change for rule_evaluations: it now upserts via `ON CONFLICT (scan_id, rule_id, resource_id) DO UPDATE` instead of delete-then-reinsert, with a scoped delete-absent pass afterward for rows no longer in the current evaluation set. A retried or replayed scan result converges on the same rows instead of a window where a concurrent reader could see zero coverage for a scan that already has some.

Findings persistence is unchanged here, still delete-then-reinsert, since the finding_key/upsert model for findings needs the schema #325 owns. This change is scoped to rule_evaluations, which is this PR's own table.

960/960 tests pass, ruff clean, single Alembic head confirmed (`3f59f83a5253`). Still in draft per m-khan's instruction. Ready whenever the merge order is confirmed.

ritiksah141:
Reviewing the updated commits (3972519, 6f18d6b, 101d47a). The upsert persistence model works correctly end to end against a migrated Postgres (idempotent replay, FAIL linked to finding_id in the same transaction, delete-absent clears stale rows), and the migration, engine wiring, and compliance fix all check out. Full suite passes and CI is green.

Three non-blocking notes, none of them holding up merge:

1. In scanner/engine.py, the evaluate()-derived FAIL finding loop calls normalize_severity without the SeverityContractError log and RULE_ERRORS_TOTAL wrapper that the scan() path has (around line 143). Both paths are fatal on an invalid severity, so behavior is consistent, but the evaluate() path skips the log and the metric. Minor observability gap.

2. scanner/rules/az_kv_006.py evaluate() calls get_key_vaults() a second time after scan() already called it, so a migrated rule issues two list calls per scan. Fine for a single reference rule, but worth keeping in mind before migrating many rules.

3. Heads-up for maintainers, not a defect in this PR: once this lands, every not-yet-migrated rule reports UNKNOWN instead of a silent PASS, so compliance scores will drop until rules are migrated. That is the intended #263 behavior, but worth announcing, and the frontend may want labels for the new unknown/error/not_applicable states.

ritiksah141 (approved):
Approving. Reviewed the updated commits (3972519, 6f18d6b, 101d47a) end to end.

Verified against a live migrated Postgres, not just mocks: the migration applies cleanly with a single Alembic head and all three CHECK constraints reject bad rows; save_scan upserts rule_evaluations idempotently (a replay converges instead of duplicating rows), links FAIL evaluations to their finding_id in the same transaction, and delete-absent removes rows no longer present; the engine records UNKNOWN for legacy rules and ERROR for evaluator exceptions, and dedupes FAIL findings against scan(); get_compliance_score reports UNKNOWN instead of a silent PASS and excludes NOT_APPLICABLE from the denominator. Full suite passes, ruff clean, CI green including DCO.

This matches the agreed reconciliation: this PR owns the evaluation contract and the evaluations upsert, while findings keep the delete-then-reinsert model because the finding_key upsert belongs to #325. Non-blocking nits are in a separate comment.

Confirming the merge order: this PR (#321) merges first as the canonical evaluation contract, then #325 rebases on top of 3f59f83a5253 and drops its duplicate rule_evaluations copy. Flagging for @m-khan-97 to sign off on the ordering so this can land.

ritiksah141 (approved):

TFT444 (approved):
Reviewed the latest commits (3972519, 6f18d6b, 101d47a). CI is 20/20 green.

The two concerns I flagged both resolved on closer inspection. The NOT IN (unnest(...)) delete-absent is safe because RuleEvaluation.__post_init__ guarantees non-null rule_id and resource_id. The SeverityContractError path aborts the entire scan anyway, so the missing ERROR row is moot.

Agreeing with the merge order: #321 first, then #325 rebases onto 3f59f83a5253.

Approving.

PR #326, opened 2026-08-30
feat(automation): finding lifecycle engine, scan outcome contracts, and pattern detection [1/5]

TFT444:
## Summary

Implements the finding lifecycle foundation described in issue #311 (Automation PR 1/5).

- **6 new DB tables** via Alembic migration `e1f2a3b4c5d6`: `scan_rule_outcomes`, `scan_lifecycle_applications`, `finding_fingerprints`, `finding_lifecycles`, `finding_lifecycle_transitions`, `patterns`
- **`LifecycleService.apply_scan()`**: idempotent, fail-closed, single transaction with `FOR UPDATE` row locking; a finding only resolves when its rule and collectors succeeded over the same authorised inventory boundary
- **`PatternService.detect_and_publish()`**: detects three pattern types (`persistent_finding`, `cross_resource_recurrence`, `reopened_finding`) with deterministic thresholds stored in the record
- **Patterns API**: `GET /api/v1/patterns` and `GET /api/v1/patterns/<id>` with tenant-scoped IDOR protection (subscription enforced in SQL, cross-subscription requests rejected 400/404)
- **Engine outcome recording**: `run_scan()` now records `SUCCESS / EMPTY_SUCCESS / PERMISSION_DENIED / TIMEOUT / FAILED` per rule; Azure HTTP 403 detected as `PERMISSION_DENIED`
- **Worker wiring**: `LifecycleService` and `PatternService` called immediately after `db.save_scan()`; lifecycle failures are non-fatal so the scan record is always preserved

## Test plan

- [ ] `tests/test_finding_lifecycle.py` — 18 unit cases covering idempotency, fail-closed behaviour, state transitions, and audit trail
- [ ] `tests/test_patterns.py` — 12 cases covering all three pattern types, IDOR regression (no-subscription-400, cross-subscription-400, cross-subscription-IDOR-404)
- [ ] Run `pytest tests/test_finding_lifecycle.py tests/test_patterns.py -v` locally (no live Postgres required; all DB calls mocked)
- [ ] Confirm `GET /api/v1/patterns` returns 400 without a subscription scope
- [ ] Confirm Alembic migration applies cleanly: `alembic upgrade head`

## Notes

This is PR 1/5 for the finding lifecycle automation epic. Subsequent PRs will add remediation-agent integration, scheduled re-scan triggers, SLA tracking, and the prod-gate controls described in #311.

### DIFF
--- a/alembic/versions/e1f2a3b4c5d6_finding_lifecycle.py
+++ b/alembic/versions/e1f2a3b4c5d6_finding_lifecycle.py
@@ -92,7 +92,7 @@ def upgrade() -> None:
             occurrence_count INTEGER NOT NULL DEFAULT 1,
             consecutive_success_count INTEGER NOT NULL DEFAULT 0,
             reopen_count INTEGER NOT NULL DEFAULT 0,
-            row_version INTEGER NOT NULL DEFAULT 0,
+            row_version INTEGER NOT NULL DEFAULT 0,  -- monotonic update counter (FOR UPDATE handles serializability)
             created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
             updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
             CONSTRAINT ck_finding_lifecycles_state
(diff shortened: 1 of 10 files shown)


m-khan-97:
@TFT444, please pause this automation PR before further implementation or review. It currently introduces `scan_rule_outcomes` and worker lifecycle writes that overlap #321’s canonical #263 evaluation contract and #325’s fenced/idempotent persistence architecture. Calling lifecycle services after `save_scan()` as non-fatal work also needs to be reconciled with #325’s atomic authoritative-write and fencing guarantees. In addition, backend CI is currently failing. We will resume #326 only after #321/#325 establish the shared schema and transaction boundary; then this PR should rebase onto those foundations and contain lifecycle/pattern behavior only, without a third scan-outcome contract.

parthrohit22 (commented):
Looked through this — the transaction/fail-closed structure in `LifecycleService.apply_scan()` is solid (idempotency sentinel inserted last, `FOR UPDATE` locking, resolution gated on `resolving_rule_ids` so a scan with no clean outcomes can't silently resolve anything), and the Patterns API's tenant/subscription scoping is done right (JWT is authoritative, query param can only narrow, both columns actually applied in the WHERE clause, not just validated).

Holding off on approve/changes since this is still draft and CI has one red test: `test_finding_lifecycle.py::TestLifecycleStateTransitions::test_reopened_finding_absent_from_success_scan_is_resolved` (`assert False`). The resolution SQL itself looks right at a glance (correctly includes `REOPENED` in the state filter), so this reads more like a mock call-count mismatch in the test than a logic bug, but worth running locally before this comes out of draft. Ping me for another look once it's green and ready.

PR #327, opened 2026-09-03
feat(scanner): add rule AZ-STOR-010 storage account missing private endpoint

### LINKED ISSUE #322: Add rule AZ‑STOR‑007 — Storage Account Missing Private Endpoint
## Rule Details
-  Rule ID: AZ-STOR-007
- Severity:  HIGH 
- Category: Storage 
- Frameworks: CIS / NIST / ISO 27001 / SOC 2

## What does it detect?
      This rule detects Azure Storage Accounts that do not have any Private Endpoint connections configured.
      A Storage Account without a Private Endpoint remains reachable over the public internet, increasing exposure to unauthorized access, data exfiltration, and network‑based attacks.

## Detection logic
- Query Storage Accounts using:
   Microsoft.Storage/storageAccounts

-Inspect the property:
  properties.privateEndpointConnections

-If:

   * The list is empty, or

   * The property is null, or

    * No Private Endpoint connection is in an Approved state

      → return one HIGH severity finding.

- If one or more Private Endpoints exist → return no findings.

## Why is this a security risk?
Storage Accounts exposed publicly can be accessed from any network on the internet.
Without Private Endpoints:

- Traffic does not stay inside the customer’s VNet

- Attack surface increases significantly

- Lateral movement becomes easier

- Compliance frameworks (CIS, NIST, ISO) are violated

- Sensitive data may be reachable from untrusted networks

Private Endpoints enforce network isolation, ensuring Storage traffic flows only through private IPs inside Azure VNets.

## Remediation steps
Azure CLI

az network private-endpoint create \
  --name <endpoint-name> \
  --resource-group <resource-group> \
  --vnet-name <vnet-name> \
  --subnet <subnet-name> \
  --private-connection-resource-id <storage-account-id> \
  --group-id blob

## Test plan
- [ ] Tested against real Azure free trial subscription
- [ ] Compliant case returns no findings
- [ ] Non-compliant case returns correct finding
- [ ] CLI playbook written and tested
- [ ] All four compliance framework JSONs updated

shariqueahmad108-ship-it:
Implements **#322** — a Storage rule that detects storage accounts reachable over the public network with **no approved Private Endpoint**.

> **Note on the rule ID:** the issue proposed `AZ-STOR-007`, but that ID is already used (Storage Account Allows TLS Below 1.2) and the storage rules currently run through `AZ-STOR-009`, so this lands as **`AZ-STOR-010`**. Happy to renumber if you'd prefer.

## Detection

For each storage account:
- If `public_network_access` is **`Disabled`** → **NOT_APPLICABLE** (already network-isolated regardless of private endpoints — no false finding).
- Else, flag (HIGH) when there is **no `privateEndpointConnections` entry whose `privateLinkServiceConnectionState.status` is `Approved`** (empty, null, or only Pending/Rejected connections).

Reads the real `azure.mgmt.storage.models` shape (`StorageAccount.private_endpoint_connections` → `PrivateEndpointConnection.private_link_service_connection_state.status`).

## Compliance mappings

| Framework | Control | |
|---|---|---|
| NIST CSF | `PR.AC-5` | Network integrity is protected |
| ISO 27001 | `A.13.1.3` | Segregation in networks |
| SOC 2 | `CC6.6` | Restricts access from outside the network boundary |
| CIS | `N/A-STOR-010` | No numbered CIS control for storage private endpoints — repo N/A convention (as AZ-STOR-006..009) |

## Tests

Four unit tests in `tests/test_rules_storage.py`, exercising **genuine SDK models** (`PrivateEndpointConnection` / `PrivateLinkServiceConnectionState`), not just `SimpleNamespace`:
- approved Private Endpoint → no finding
- `public_network_access=Disabled` → NOT_APPLICABLE
- no Private Endpoint, public reachable → one HIGH finding
- only a Pending connection → flagged (not Approved)

```
pytest tests/test_rules_storage.py   -> 23 passed
ruff check .            -> clean
ruff format --check .   -> clean
```

## Honest testing note

I don't have an Azure subscription, so I couldn't run this against a live account. The detection is fully covered by the unit tests above, which instantiate real `azure-mgmt-storage` models so the attribute shape (`private_endpoint_connections`, `private_link_service_connection_state.status`) is validated against the pinned SDK rather than invented.

Includes the remediation playbook (`fix_az_stor_010.sh`) with guarded args.

Resolves #322.

### DIFF
--- a/compliance/frameworks/cis_azure_benchmark.json
+++ b/compliance/frameworks/cis_azure_benchmark.json
@@ -33,6 +33,11 @@
       "control_name": "Required Blob Container Immutability Missing",
       "description": "OpenShield checks this service-specific control without claiming an unrelated CIS recommendation."
     },
+    "AZ-STOR-010": {
+      "control_id": "N/A-STOR-010",
+      "control_name": "Storage Account Missing Private Endpoint",
+      "description": "OpenShield checks this service-specific control without claiming an unrelated CIS recommendation."
+    },
     "AZ-DB-005": {"control_id": "N/A-DB-005", "control_name": "SQL Server Microsoft Entra-Only Authentication Not Enforced", "description": "Service-specific OpenShield control."},
     "AZ-DB-006": {"control_id": "N/A-DB-006", "control_name": "SQL Vulnerability Assessment Not Configured", "description": "Service-specific OpenShield control."},
     "AZ-DB-007": {"control_id": "N/A-DB-007", "control_name": "SQL Auditing Retention Below Minimum", "description": "Service-specific OpenShield control."},
--- a/compliance/frameworks/iso27001.json
+++ b/compliance/frameworks/iso27001.json
@@ -33,6 +33,11 @@
       "control_name": "Required Blob Container Immutability Missing",
       "description": "OpenShield checks this service-specific control without claiming an unrelated ISO 27001 recommendation."
     },
+    "AZ-STOR-010": {
+      "control_id": "A.13.1.3",
+      "control_name": "Segregation in networks",
+      "description": "A storage account reachable over the public network has no approved Private Endpoint, leaving its blob/file/queue/table endpoints on the public internet. A.13.1.3 requires segregation of networks; a Private Endpoint isolates storage traffic within the private VNet."
+    },
     "AZ-DB-005": {"control_id": "A.9.4.2", "control_name": "Secure log-on procedures", "description": "SQL authentication is restricted to approved Entra identities."},
     "AZ-DB-006": {"control_id": "A.12.6.1", "control_name": "Management of technical vulnerabilities", "description": "Required SQL vulnerability assessment is configured."},
     "AZ-DB-007": {"control_id": "A.12.4.1", "control_name": "Event logging", "description": "SQL audit logs are retained according to policy."},
--- a/compliance/frameworks/nist_csf.json
+++ b/compliance/frameworks/nist_csf.json
@@ -33,6 +33,11 @@
       "control_name": "Required Blob Container Immutability Missing",
       "description": "OpenShield checks this service-specific control without claiming an unrelated NIST recommendation."
     },
+    "AZ-STOR-010": {
+      "control_id": "PR.AC-5",
+      "control_name": "Network integrity is protected",
+      "description": "A storage account reachable over the public network has no approved Private Endpoint, so its data-plane endpoints stay internet-reachable. PR.AC-5 requires that network integrity is protected; a Private Endpoint keeps storage traffic on private IPs inside the VNet."
+    },
     "AZ-DB-005": {"control_id": "PR.AC-6", "control_name": "Identity proofing and authentication", "description": "SQL authentication is restricted to approved Entra identities."},
     "AZ-DB-006": {"control_id": "DE.CM-8", "control_name": "Vulnerability scans are performed", "description": "Required SQL vulnerability assessment is configured."},
     "AZ-DB-007": {"control_id": "PR.PT-1", "control_name": "Audit/log records are determined, documented, implemented, and reviewed", "description": "SQL audit logs are retained for at least 90 days in accordance with the defined audit policy."},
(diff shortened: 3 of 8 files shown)


TFT444 (changes requested):
The rule logic, SDK attribute paths, enum handling, playbook, and compliance mappings are all correct. The four tests cover the main paths well. Two issues need fixing before merge.

---

## Bug 1  No test for `private_endpoint_connections = None` (potential false positive in production)

The rule's `_has_approved_private_endpoint` helper does:

```python
for connection in getattr(account, "private_endpoint_connections", None) or []:
```

The `or []` means a `None` value (what `StorageAccount.__init__` sets by default before the SDK populates it from the API response) is silently treated as "no approved endpoint," flagging the account as HIGH. That is the safe direction for a security rule, but if `get_storage_accounts()` returns real SDK objects whose PE list was not populated (e.g., because the caller lacks `Microsoft.Storage/storageAccounts/privateEndpointConnections/read` permission), every such account gets a false HIGH finding.

The fix is not to change the rule logic but to:
1. Add a test asserting the current behavior so it is documented and intentional.
2. Add a `logger.warning` for the `None` case (consistent with how `az_stor_009.py` handles `None` container lists and `az_sc_002.py` handles unknown `public_network_access`), so operators can distinguish a real absence of PEs from a permissions gap.

---

## Bug 2 `docs/rules-reference.md` not updated

Every existing storage rule through AZ-STOR-009 has a row in `docs/rules-reference.md`. AZ-STOR-010 has no row. The engine auto-discovers rules via glob so the rule runs, but the reference doc is the human-readable registry rendered on the learn page. Please add the missing row.

---

## Observation (not blocking, pre-existing) stale control ID in `nist_csf.json`

The AZ-STOR-010 NIST CSF entry in the diff is correct (`PR.AC-5`), but a nearby pre-existing entry for AZ-DB-007 in the same file uses `"control_id": "A.12.4.1"` which is an ISO 27001 ID in a NIST file. Not introduced by this PR and not a blocker here, but flagging since the author touched this file.

---

## Verified correct

Detection field path (`private_endpoint_connections` → `private_link_service_connection_state.status`) is correct per the pinned Azure SDK. `enum_str` handles both real SDK enums and plain strings correctly. No RULE_ID collision. Playbook (`fix_az_stor_010.sh`) exists, uses `set -euo pipefail`, validates args, correctly creates the PE then disables public network access. All 4 compliance frameworks updated with accurate control mappings. NOT_APPLICABLE handling for `public_network_access == "Disabled"` accounts is correct.

parthrohit22 (changes requested):
I reviewed the rule implementation, Azure SDK usage, compliance mappings, remediation playbook, and tests.

The core detection logic looks correct, but two issues remain: unavailable private-endpoint evidence is silently treated as confirmed absence, and `AZ-STOR-010` is missing from the rules reference documentation.

Please address both issues and add the corresponding regression coverage before requesting another review.

TFT444 (approved):
Re-reviewed. Both issues resolved: private_endpoint_connections = None is now treated as indeterminate (UNKNOWN) with a log warning and a regression test, and AZ-STOR-010 is documented in rules-reference.md. Approved.

TFT444 (approved):
Both blockers from the prior reviews are resolved: the `private_endpoint_connections = None` path now returns UNKNOWN with a `logger.warning` and has a regression test, and AZ-STOR-010 is documented in `docs/rules-reference.md`. CI is green on the approved commit. Approving.

TFT444:
@parthrohit22 @ritiksah141  can you re-review please its good to go from my side

parthrohit22 (approved):
Re-reviewed `94ea7cf`. Both items from my 6 Sep review are addressed — approving.

**1. Unavailable evidence is no longer treated as confirmed absence.** `private_endpoint_connections is None` now logs and `continue`s as indeterminate, and only a genuine empty list (or a list with no Approved connection) produces a finding. The docstring states the distinction explicitly, which matters here because the next reader's instinct is to collapse `None` and `[]` back together. `test_stor_010_unavailable_evidence_is_indeterminate_not_flagged` pins it, so a regression fails CI rather than quietly resurfacing a false HIGH on an account whose connections simply weren't populated.

Treating `public_network_access: Disabled` as NOT_APPLICABLE before the private-endpoint check is also the right ordering — an account closed to the public network isn't exposed regardless of how it got there.

**2. Documentation.** `AZ-STOR-010` is in `docs/rules-reference.md` with its severity, category and framework mappings.

CI green, remediation playbook present and named per convention. Thanks for the quick turnaround on both.

Issue #304, opened 2026-08-21
infra: establish one reproducible enterprise deployment and operations baseline


m-khan-97:
## What is missing

Render configuration, Terraform, Docker and Compose describe different systems. Terraform omits staging and the worker, contains placeholder remote-state configuration, and has no committed provider lock. The deployed RAG vector store is neither built nor mounted. There are no project SLOs, alert ownership, RTO/RPO, restore drill, capacity evidence or tested coordinated rollback.

Runtime dependencies are also resolved from broad/unhashed requirements, while release images are published independently of the strongest signed source-release gate.

## Required work

Choose one authoritative deployment topology and make it reproducible, observable and recoverable.

## Acceptance criteria

- Live resources are imported into one approved IaC source of truth with locked state/providers and drift checks.
- Production and isolated staging include API, worker, database, backups, alerts and required data artifacts.
- RAG is a versioned verified artifact/managed store or its unavailable features are disabled.
- Runtime/dev dependencies are separated and locked with hashes; vulnerability exceptions have owner, issue and expiry.
- Container digest is tied to a verified main-ancestor release tag, scanned, SBOMed, attested and signed.
- SLOs, alert routes, on-call ownership, capacity limits, retention, RTO and RPO are approved.
- Backup restore, expand/contract migration and coordinated rollback drills are executed and recorded.
- API and worker metrics are observable from the deployed topology.

Related: #199 and #243.

m-khan-97:
I am picking up the Python dependency-reproducibility slice of this issue: separate runtime and test inputs, generate fully pinned hash-checked locks for the supported runtime, and verify their use in CI and the container install. This does not take over deployment topology, cloud changes or the wider operations work. I will keep those out of this PR and leave #304 open for the remaining acceptance criteria.

m-khan-97:
The dependency-reproducibility slice is now ready for review in #336. It separates runtime/development inputs, adds pinned hash-checked locks, and verifies them through the existing CI gate. Local validation: 994 tests passed, 86.50% coverage, clean runtime/development installs, and no known vulnerabilities from the runtime audit under the existing policy. Ritik and Vishnu have been requested for review; CI is running. This is only one slice of #304, so the issue remains open for the deployment, release and operations work.

m-khan-97:
[...]

PR #336, opened 2026-09-08
build: lock Python runtime and development dependencies with hashes

m-khan-97:
## What changed

I am taking the dependency-reproducibility slice of #304. Production currently installs test tools and resolves transitive dependencies afresh; this makes a successful CI run less useful as evidence for the next deployment.

- Separate direct runtime, development and lock-tool inputs.
- Commit fully pinned SHA-256 locks, with development constrained to the production package versions.
- Require hashes and binary wheels in the container, native Render build commands, deployment workflow and relevant CI installs.
- Verify input/lock consistency within the existing rule-validation gate. The check operates in a temporary directory and preserves compatible pins rather than upgrading on every CI run.
- Add 12 regression checks, including offline good/bad wheel hashes, stale-lock handling, runtime/dev separation and CI wiring.
- Document the update process and declare Python 3.11 as the repository's native-build default, matching CI and Docker.

Existing exact direct runtime pins are preserved. Previously open ranges and transitive dependencies now have explicit versions. No scanner features, live cloud settings, review-policy changes or new vulnerability exceptions are included.

## Validation

- Clean Python 3.11/Linux development installation with `--require-hashes --only-binary=:all:` and `pip check` passed.
- Separate clean runtime installation and API/scanner imports passed; pytest, coverage, ruff and pip-tools are absent from that environment.
- The runtime environment also built the container's BM25 artifact successfully (984 chunks). A subsequent targeted hallucination-guard run passed 2 tests and skipped 2 requiring an AI API key; no live LLM was called.
- Input/lock consistency check passed.
- Full local pytest suite: **994 passed, 9 skipped; 86.50% API/scanner coverage**, above the existing 80% gate. Skips were three missing-BM25-index tests and six PostgreSQL-dependent cases. One existing pytest iterator-deprecation warning remains.
- All 12 dependency-lock tests passed, including rejection of a mismatched wheel hash without network access.
- Repository-wide ruff check and format check passed; `git diff --check` passed.
- Runtime pip-audit with the existing CI exception arguments reported no known vulnerabilities. No exceptions were added or relaxed.

## Limits and review notes

Docker is unavailable locally, so the existing container runtime/scan CI job must validate the image. Live Azure, PostgreSQL integration and deployment were not exercised locally. Before deployment, verify that a Render dashboard `PYTHON_VERSION` does not override `.python-version`.

This does not claim a fully reproducible container: base-image/OS packages and the existing pip/setuptools/wheel bootstrap remain separate work. Security-scanner tooling also retains its existing independent environments.

Refs #304 — intentionally not closing it. Deployment topology, release provenance, backup/restore drills and operations acceptance criteria remain open.

## Dependencies

Depends-On: none


m-khan-97:
CI caught an integration mistake in my lock-check setup: creating .lock-tools inside the checkout caused the existing credential scan to inspect pip/setuptools placeholder strings. I have moved the tool environment into RUNNER_TEMP and added a regression assertion for that boundary. The credential scan and its patterns are unchanged. All 19 focused dependency/container-configuration tests and lint/format checks pass locally. The new commit triggers a fresh CI run.

parthrohit22 (approved):
Reviewed `9358488`. The three-tier split (runtime / dev constrained to `-c requirements.txt` / isolated lock tooling in `requirements-lock.*`) is the right shape, and the enforcement is consistent across the surfaces that matter: `--require-hashes --only-binary=:all:` in the Dockerfile, all four `render.yaml` build commands, the CI installs, and `deploy.yml`, each followed by `pip check`.

Verified locally:

- `requirements.txt` pins 66 packages, every one carrying `--hash=sha256:` lines; every direct dependency in `requirements.in` is present as an `==` pin in the lock.
- All 12 `tests/test_dependency_locks.py` pass, including the offline mismatched-wheel-hash rejection.
- `scripts/lock_dependencies.py --check` seeds the temp resolve with the committed `.txt` files, so only a genuine input change (not a new upstream release) makes the gate fail — that's the usual pip-tools-in-CI failure mode handled well.
- All 21 PR checks are green, including Container Scan (Trivy), SCA (pip-audit) with the existing exception args, and Rule & Compliance Validation (which runs the lock check).

Non-blocking:

1. The `Install ruff` lint step now installs the full `requirements-dev.txt` (runtime + ~40 Azure SDKs + pytest) under `--require-hashes` just to get `ruff`. It works, but a `ruff`-only pinned+hashed input would keep the lint job lean and independent of the runtime resolution.
2. `lock_dependencies.py --check` still needs PyPI reachable to re-resolve changed inputs; a registry outage fails that step even when the locks are correct. The seed-with-committed-locks logic keeps this to genuine input diffs, so acceptable — a one-line note in the CI step would set expectations.
3. Dockerfile's `pip==26.1.2 setuptools==83.0.0 wheel==0.46.3` bootstrap is still unhashed. The PR body scopes this out explicitly; flagging only so it's a recorded boundary.
4. `.python-version` (3.11) is enforced only by `lock_dependencies.py`, and only for lock generation. The Dockerfile `FROM` and CI `setup-python` still name their versions independently. Consistent today; an assertion that the Dockerfile base tracks `.python-version` would harden against future drift.

None of these block. Approving.

ritiksah141 (approved):

ritiksah141 (approved):

m-khan-97:
@Vishnu2707, could you take the final review when you have a moment? Parth and Ritik have approved the current head, the checks have no failures, and there are no unresolved review threads. GitHub still marks merging as blocked, so I am leaving it unmerged and not bypassing any protections. The scope and remaining #304 work are recorded in the PR.

Vishnu2707 (approved):
Approved, LGTM.

TFT444 (approved):
--require-hashes --only-binary=:all: is enforced consistently across CI jobs, Dockerfile, and Render build commands. The three-layer split (runtime / dev / lock-tool) is clean and the lock script documents the update workflow. Approved.

